Issue #6, opened 2017-02-01
Add support for the upcoming Debian Stretch


zelda14:
While installing **ansible-openwisp2** on **Debian Stretch RC1** I face an *uwsgi* crash as ansible fetched uwsgi 2.0.14 from the PyPi repo. The crash is related to an *OpenSSL 1.1* incompatibility.

However while digging and experimenting, Debian successfully installed uwsgi from its own repo. 
In order to have a functionnal installation I modified the file : 

`/etc/ansible/roles/openwisp.openwisp2/tasks/pip.yml`

with the following informations :

```
- name: Install uwsgi
  pip:
    name: http://http.debian.net/debian/pool/main/u/uwsgi/uwsgi_2.0.14+20170111.orig.tar.gz
    state: latest
    virtualenv: "{{ virtualenv_path }}"
    virtualenv_python: "{{ openwisp2_python }}"
    virtualenv_site_packages: yes
  with_items:
    - uwsgi
  notify: reload supervisor
```

That way pip fetch, build and doesn't complain.

Thanks

nemesisdesign:
More info supplied on the [mailing list](https://groups.google.com/d/msg/openwisp/ggPqa9e1PVQ/MkBczcH8EAAJ):

I've updated my configuration and got some issues I solved, here is my feedback on the upcoming Debian stretch.

**1**: no issue while running:  ansible-galaxy install --force openwisp.openwisp2

**2**: Python 2 and Python 3 are installed on my system.

when running:

    ansible-playbook -i hosts playbook.yml -u xavier -k --ask-sudo-pass

I got a problem with:

    "No package matching 'python-software-properties' is available"

In my environment the package has been deprecated in favor of:

    sudo apt-get install python-software-properties
    Reading package lists... Done 
    Building dependency tree
    Reading state information... Done 
    Package python-software-properties is not available, but is referred to by another package. 
    This may mean that the package is missing, has been obsoleted, or 
    is only available from another source 
    However the following packages replace it: 
        software-properties-common

I updated the `apt.yml` accordingly and the problem is gone

**3**: - Then I face:

   ImportError: No module named 'pyparsing'

There is two ways to fix it.
Add the dependency in `apt.yml`:

```yaml
- name: Install python3 packages
  when: openwisp2_python == "python3"
  apt: name={{ item }} state=latest
  with_items:
    - python3
    - python3-pip
    - python3-dev
    - python-virtualenv
    - python3-pyparsing
````

[...]

PR #8, opened 2017-03-12
Add support for debian 9

ritwickdsouza:
References #6 . Still needs to be verified.

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -19,7 +19,7 @@
 # fixes issue described in https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user
 - name: Install acl if acting as non-root user
   apt: name=acl state=latest
-  when: ansible_user != 'root'
+  when: ansible_user is not defined or ansible_user != 'root'
 
 - name: ensure supervisor is started
   service: name=supervisor state=started


tasks/pip.yml
```diff
@@ -58,7 +58,7 @@
 
 - name: Install uwsgi
   pip:
-    name: "{{ item }}"
+    name: http://http.debian.net/debian/pool/main/u/uwsgi/uwsgi_2.0.14+20170111.orig.tar.gz
```
nemesisdesign: this approach is bad. `{{ item }}` represents a value that is set below.
You should determine this value conditionally, if it's debian stretch, insert this new string, otherwise keep the old package name.

tasks/apt.yml
```diff
@@ -40,6 +41,8 @@
     - python3-pip
     - python3-dev
     - python-virtualenv
+    - python3-pyparsing
+    - python3-software-properties
```
nemesisdesign: Are you sure these two packages are present on all the supported distributions? I think it would be better to install `python3-pyparsing` and `python3-software-properties` only if the current distribution is debian stretch.

ritwickdsouza: It's believe its supported by Ubuntu and Debian, but then again scaling it in the future would require to make changes. So your way would be a better approach.

tasks/pip.yml
ritwickdsouza: Should i create a new section just for debian or is there a shorter/better way to make such a change ?

nemesisdesign: I'm not entirely convinced of this bug. Can you create a debian stretch VM with virtualbox or anything you like and test the playbook without this change (leave this line how it was before) and let me know what happens?

ritwickdsouza:
@nemesisdesign Sorry for the delay, was caught up with other work.
So i tried to install the master branch on Debian Stretch RC 2 (had some issues installing RC 1) didn't face the uwsgi. But while running a playbook the `"No package matching 'python-software-properties' is available"` still exists. So I'll just work on that

ritwickdsouza:
I get the following when i install it on Debian Stretch RC1. 
`[WARNING]: - Stouts.postfix was NOT installed successfully: error executing: git clone https://github.com/nemesisdesign/Stouts.postfix Stouts.postfix`
And if i ignore it and try to run a playbook, i get the following error
```ERROR! the role 'Stouts.postfix' was not found in /root/roles:/etc/ansible/roles:/etc/ansible/roles:/root

The error appears to have been in '/etc/ansible/roles/openwisp.openwisp2/meta/main.yml': line 4, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

dependencies:
    - src: https://github.com/nemesisdesign/Stouts.postfix
      ^ here```
Is it a problem with 'Stouts' as it's not tested on distributions other than Ubuntu ?

**Comment by nemesisdesign** [comment]
@ritwickdsouza do you get that clone error when installing the role or when? What ansible version are you using?

**Comment by ritwickdsouza** [comment]
I got this warning when installing role for the first time.
```
[WARNING]: - Stouts.postfix was NOT installed successfully: error executing: git clone https://github.com/nemesisdesign/Stouts.postfix Stouts.postfix
```
If I ignore the warning and run a playbook i get the clone error.
I'm using ansible 2.2.1.0 on Debian Stretch RC1

**Comment by nemesisdesign** [comment]
@ritwickdsouza Is git installed on the machine where you install ansible to run the playbook?

**Comment by ritwickdsouza** [comment]
@nemesisdesign My bad. Forgot to install git. :P

File: tasks/apt.yml [hunk 3]
```diff
@@ -54,3 +53,21 @@
...
+
+- name: Install python-software-properties
+  when: ansible_distribution =! 'Debian'
+  apt: name={{ item }} state=latest
+  with_items:
+    - python-software-properties
+
+- name: Install git if using development version of any openwisp module
```
nemesisdesign: you can remove this task because it's not present anymore on master

tasks/apt.yml
```diff
@@ -54,3 +53,21 @@
...
+- name: Install python3-pyparsing and python3-software-properties
+  ignore_errors: yes
+  when: ansible_distribution == 'Debian' and ansible_distribution_version == '9'
+  apt: name={{ item }} state=latest
+  with_items:
+    - python3-pyparsing
+    - python3-software-properties
+
+- name: Install python-software-properties
+  when: ansible_distribution =! 'Debian'
```
nemesisdesign: I think the `when` clause needs to be:

```
when: ansible_distribution == 'Debian' and ansible_distribution_version == '9'
```

What do you think?

tasks/apt.yml
```diff
@@ -6,7 +6,7 @@
   notify: reload systemd
   with_items:
     - sudo
-    - python-software-properties
+    - software-properties-common
```
ritwickdsouza: @nemesisdesign  `python-software-properties` is deprecated so replaced it with the new `software-properties-common`

tasks/apt.yml
```diff
@@ -41,6 +41,12 @@
     - python3-dev
     - python-virtualenv
 
+- name: Install python3-pyparsing
```
ritwickdsouza: @nemesisdesign  I believe this is sufficient. What do you think ?

ritwickdsouza:
Created new PR #10

Issue #5, opened 2017-01-27
Add support for Fedora / RedHat / Centos


nemesisdesign:
Fedora is a popular linux distribution that is suitable to be used for hosting an openwisp2 instance.

The parts to change to add support to Fedora are the following ones:

- management of system packages: use yum module instead of apt and the names of some packages are different
- nginx configuration, path differ slightly
- supervisord configuration, a few things differ here too

moonwolf-github:
And CentOS7/RHEL7 support is also missing.

moonwolf-github:
I think i did it. It's not yet flawlessly nor pretty (it's my first contact with ansible and yaml) but looks like it works...

nemesisdesign:
@moonwolf-github show me something and I'll try to help you out

moonwolf-github:
I've forked your repo and my work is here: https://github.com/moonwolf-github/ansible-openwisp2/tree/centos_support. AFAIR there were two main problems: during virtualenv creation one python file was missing, and i had to turn off SELinux because of Django app path (/opt/). i didn't want to mess around with file locations. Both issues are somehow fixable of course.
I'll verify this out tomorrow and report back.

nemesisdesign:
Hi @moonwolf-github, thank you for this update.

moonwolf-github:
As i said there is a problem with virtualenv creation:
`{"cmd": "/usr/bin/virtualenv --system-site-packages -ppython3 /opt/openwisp2/env", "failed": true, "item": "pip", "msg": "stdout: Using base prefix '/usr'\nNew python executable in /opt/openwisp2/env/bin/python3\nAlso creating executable in /opt/openwisp2/env/bin/python\nERROR: The executable /opt/openwisp2/env/bin/python3 is not functioning\nERROR: It thinks sys.prefix is '/tmp' (should be '/opt/openwisp2/env')\nERROR: virtualenv is not compatible with this system or executable\nRunning virtualenv with interpreter /usr/bin/python3\n\n:stderr: Failed to import the site module\nTraceback (most recent call last):\n  File \"/opt/openwisp2/env/lib64/python3.4/site.py\", line 67, in <module>\n    import os\n  File \"/opt/openwisp2/env/lib64/python3.4/os.py\", line 618, in <module>\n    from _collections_abc import MutableMapping\nImportError: No module named '_collections_abc'\n"}`

[...]

PR #11, opened 2017-03-22
Redhat & Centos support

moonwolf-github:
This version at leas works flawlessly - at least on CentOS7 "Minimal install". But it has numerous issues i was unable to resolve:
* "updating yum cache" means updating whole system (this is easy fixable i guess, but it may be better to leave it?)
* i had to install python2's pip despite python3 is in use
* many `when` directives
* separate `django.yml` file for centos (because of group is different)
* worst of all - i had to disable SELinux (because ngnix logs are in different place than default)

@nemesisdesign  - what do you think? How can i improve this further?

References #5.


nemesisdesign:
Questions:

- does updating the yum cache automatically upgrades also the packages as well? Or is it just the equivalent of `apt-get update`?
- could you summarize why python2 instead of python3?
- the different group used in `django_centos.yml` could become a variable and then you could use a conditional just to set the variable on centos/redhat/fedora, this would avoid the duplication of all the commands related to django

moonwolf-github:
* as i understand, now it upgrades packages also. AFAIR yum doesn't need refreshing cache as it gets refreshed as needed. Unless ansible does this differently. I'll investigate this tomorrow
* because `pip`  task expects `pip` or `pip2` binary and this is provided by `python2-pip` package. Python3 provides `pip3` binary.
* ok, i'll try to do this

Thanks for info.

nemesisdesign:
There must be a way to solve the `pip` issue related to python3, I guess we are not the first ones encountering this issue. It's important to solve it because [django 2.0 will be python3 only](https://news.ycombinator.com/item?id=13433927) and I want openwisp2 to move to django 2.0 when it will be released.

moonwolf-github:
(push --force is a nasty thing, but merge commits also are)

I had to link `/usr/bin/pip3` to `/usr/bin/pip`. I don't know if there's smarter way to do this.
This way it break things if someone will try to install python2's `pip`. But `virtualenv` is broken same way (it's been installed with `pip`).

nemesisdesign:
I'm dedicating some time to this now, I will leave some more comments to suggest improvements and in the meanwhile I will setup a new VM that I will use for testing this.

nemesisdesign:
I think we could also use `ansible_os_family` in some cases, here's the ansible code that defines the families:

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts.py#L675-L685

```python
OS_FAMILY = dict(
   RedHat = 'RedHat', Fedora = 'RedHat', CentOS = 'RedHat', Scientific = 'RedHat',
   SLC = 'RedHat', Ascendos = 'RedHat', CloudLinux = 'RedHat', PSBM = 'RedHat',
   OracleLinux = 'RedHat', OVS = 'RedHat', OEL = 'RedHat', Amazon = 'RedHat',
   XenServer = 'RedHat', Ubuntu = 'Debian', Debian = 'Debian', Raspbian = 'Debian', Slackware = 'Slackware', SLES = 'Suse',
   SLED = 'Suse', openSUSE = 'Suse', openSUSE_Tumbleweed = 'Suse', SuSE = 'Suse', SLES_SAP = 'Suse', SUSE_LINUX = 'Suse', Gentoo = 'Gentoo',
   Funtoo = 'Gentoo', Archlinux = 'Archlinux', Manjaro = 'Archlinux', Mandriva = 'Mandrake', Mandrake = 'Mandrake', Altlinux = 'Altlinux', SMGL = 'SMGL',
   Solaris = 'Solaris', Nexenta = 'Solaris', OmniOS = 'Solaris', OpenIndiana = 'Solaris',
   SmartOS = 'Solaris', AIX = 'AIX', Alpine = 'Alpine', MacOSX = 'Darwin',
   FreeBSD = 'FreeBSD', HPUX = 'HP-UX', openSUSE_Leap = 'Suse', Neon = 'Debian'
)
```

moonwolf-github:
Good point.

tasks/selinux_centos.yml
```diff
@@ -0,0 +1,10 @@
+- name: Set http_can_network_connect
+  seboolean:
+    name: httpd_can_network_connect
+    state: yes
+    persistent: yes
+
+# Disable SELinux
+- name: Disable SELinux (shouldn't be done!)
+  selinux:
+    state: disabled
```
nemesisdesign: Regarding this problem, @dariomas suggested to set selinux in "permissive mode", see what kind of suggestions you get from selinux in its logs and add the necessary rules, do you think it's possible?

tasks/nginx.yml
```diff
@@ -18,10 +18,19 @@
...
     src: "/etc/nginx/sites-available/{{ inventory_hostname }}"
     dest: "/etc/nginx/sites-enabled/{{ inventory_hostname }}"
     state: link
   notify: restart nginx
+  when: ansible_distribution != 'CentOS' and ansible_distribution != 'Red Hat Enterprise Linux'
+
+- name: nginx site
+  template:
+    src: ../templates/nginx.j2
+    dest: "/etc/nginx/conf.d/{{ inventory_hostname }}.conf"
```
nemesisdesign: It looks like only the filesystem path changes, if I'm not wrong, the template used is the same.

You could use the same approach I suggested for `www_group` and `www_data`, set these variables using `set_fact` depending on `ansible_os_family`

tasks/supervisor.yml
```diff
@@ -12,3 +12,21 @@
...
+  action: template src=../templates/supervisor.j2 dest=/etc/supervisord.d/openwisp2.ini
+  notify: reload supervisor
+  tags:
+    - openwisp2
+    - supervisor
+  when: ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux'
+
+- name: supervisor uwsgi modification for Redhat/CentOS
+  shell: >
+    grep -v "user=" /etc/supervisord.d/openwisp2.ini > /tmp/tempconf.tmp && echo "user=nginx" >> /tmp/tempconf.tmp && rm -f /etc/supervisord.d/openwisp2.ini && mv /tmp/tempconf.tmp /etc/supervisord.d/openwisp2.ini
```
nemesisdesign: Here it would be much better to change the [supervisor template (at line 2)](https://github.com/openwisp/ansible-openwisp2/blob/master/templates/supervisor.j2#L2) so that it would use the variable `{{ www_user }}` instead of the current hardcoded value.

If you do that you won't need to perform this task at all.

tasks/yum_centos.yml
```diff
@@ -0,0 +1,61 @@
+- name: Update packages
```
nemesisdesign: @dariomas confirmed this is not needed so yeah let's remove this task 👍

tasks/selinux_centos.yml
moonwolf-github: It complained about nginx log files. They are type `usr_t` and when i allow nginx access to this type i get warning it is a generic type. I suppose many files are of this type so allowing access is almost as disabling SELinux.

tasks/nginx.yml
moonwolf-github: Unfortunately in case of Debian it's a link to file in `sites-enabled`. In RedHats this is plain file. So variables won't eliminate task for RedHat.

moonwolf-github:
OK, should be better now. We still have nginx and SELinux issues. Only solution to get rid of SELinux problems is move log files to `/var/log/nginx`, i think. I don't know a way to relabel files with ansible.

nemesisdesign:
@moonwolf-github yes it looks better. I intend to merge it on a local branch, test it and improve it a bit more, but I don't have time right now because I'm preparing for a couple of conferences which are happening next week. I will likely have time to do this from the 10th of April.

One question: are you using this path to manage your own instance?

moonwolf-github:
@nemesisdesign i've deployed it using very early version. Then i was testing changes on clean VM.

Would you like me to squash commits or edit log messages (i noticed there are some kind of tags in logs)?

nemesisdesign:
@moonwolf-github yes please, squashing would be good.

Regarding the tags, you mean the prefixes I use in commit messages? You could use something like:

`[redhat] Added initial support for Redhat/Centos/Fedora #5`

I've also opened a new thread on the mailing list for this patch to ask for help:
https://groups.google.com/d/msg/openwisp/CHU1xmNK4BY/yAwRpAomCAAJ

I've also asked on social media:
- [facebook](https://www.facebook.com/OpenWISP/photos/a.849982858444124.1073741827.240232962752453/1179973565445050/?type=3&theater)
- [twitter](https://twitter.com/openWISP/status/847371457222946817)

Germano0:
I asked for suggestions in #selinux Freenode IRC channel:
```
<Zhenech> germano, the usual aproach would be to add an alias file context to treat your finny log folder the same as /var/log/nginx and relabel
<Zhenech> germano, https://linux.die.net/man/8/semanage -- "Substitute /home1 with /home when setting file context
<Zhenech> mhh, http://man7.org/linux/man-pages/man8/semanage-fcontext.8.html is better formated
```

moonwolf-github:
Ansible's SeLinux module can't do this AFAIK. So, shell execution?

moonwolf-github:
One important thing: this patch doesn't support Fedora (yet?). I'm testing this on F25 Server right now.

Germano0:
@moonwolf-github
```
<sfix>you can just use the type directly
<sfix> httpd_var_log_t
```

evgeni:
Zhenech here.

Yeah. The ansible `sefcontext` module does not seem to handle equals. Here is a yak to shave for you ;-)
But yeah, otherwise either use `command` (ugh) or `sefcontext` but duplicate the definition of the folder and set the type to `httpd_var_log_t`.

nemesisdesign:
thanks a lot to @Germano0 and @evgeni for helping us out.

A `command` task would be good enough.

@moonwolf-github: I haven't been able to try it on Fedora because I had an issue in setting up a fedora VM in a proxomox instance I have access to

moonwolf-github:
This almost works good. I don't like this much because of this:
```
[root@localhost ~]# ll -Z /opt/openwisp2/log
-rw-r--r--. nginx nginx unconfined_u:object_r:var_log_t:s0 error.log
-rw-r--r--. root  root  system_u:object_r:httpd_log_t:s0 nginx.access.log
-rw-r--r--. root  root  system_u:object_r:httpd_log_t:s0 nginx.error.log
-rw-r--r--. root  root  system_u:object_r:var_log_t:s0   uwsgi.log
```
nginx files are created later with nginx start that's why have other labels. Will look into it further.

I'll hold squashing until figure out this issue.

nemesisdesign:
@moonwolf-github great news, thx for updating us

moonwolf-github:
I think that's better. But because SELinux for me is some kind of magic, somebody should look at it, as this is matter of security.
SElinux is now in permissive mode - for testing purposes only.

moonwolf-github:
OK, i just found a bug - nginx doesn't start after reboot.

tasks/yum.yml
```diff
@@ -21,7 +21,10 @@
     - policycoreutils-python
 
 - name: ensure supervisor is started
-  service: name=supervisord state=started
+  service: name=supervisord state=started enabled=yes
```
nemesisdesign: so by default these two are not started at boot? Anyway adding that doesn't hurt

moonwolf-github: Both say `vendor preset: disabled` so i assumed both are not started.

nemesisdesign:
@ritwickdsouza do you have any experience with Fedora, Centos or Redhat?

ritwickdsouza:
CentOS would be good. I will work on that.

nemesisdesign:
@moonwolf-github has done a lot of work. If you could test his patch and try to improve the solution regarding python-virtualenv it would be very helpful

tasks/pip.yml
```diff
@@ -1,3 +1,11 @@
+- name: Install virtualenv with pip
```
nemesisdesign: @moonwolf-github & @ritwickdsouza thinking about this virtualenv issue, have you thought about specifying the python version contained in `openwisp2_python` here and removing the symbolic link later?

nemesisdesign:
@moonwolf-github could you squash the commits into one? I'd like to merge the patch. Let me know!

nemesisdesign:
Done. Thank you all @moonwolf-github @Germano0 @evgeni

moonwolf-github:
Sorry, i wasn't around for last few days. Glad to see it all worked out!

PR #13, opened 2017-04-27
added nginx http2 config option; default: false

netminion:
I added a new nginx server option for http2 in nginx template and configuration option in defaults/main.yml
The default setting is false


nemesisdesign:
@netminion thanks for contributing.

If you are interested in general discussions or need to ask questions, we also have a mailing list or an IRC channel: http://openwisp.org/support.html

Issue #20, opened 2017-07-12
[network-topology] Add cron to run `update_topology` command


nemesisdesign:
#19 introduced the openwisp-network-topology module developed by @rohithasrk but we forgot to add the cronjob to run `./manage.py update_topology` which is needed to automatically update the topologies using the FETCH strategy every `x` minutes.

When `openwisp2_network_topology` is set to `true` we should add a cron that runs the `update_topology` command, eg:

```
3 * * * * /root/.virtualenvs/ninux_graph/bin/python /var/www/ninux_graph/manage.py update_topology
```

In order to control the cron frequency, we could copy the syntax used by the role [finn.letsencrypt](https://github.com/thefinn93/ansible-letsencrypt/blob/master/tasks/main.yml#L102-L109) (which [we suggest in our README to handle the SSL cert](https://github.com/openwisp/ansible-openwisp2#automatic-ssl-certificate)).

rohithasrk:
This shall be added once the PRs in **django-netjsongraph** and **openwisp-network-topology** are merged.

nemesisdesign:
Closed by #21

PR #21, opened 2017-07-20
[WIP] Added cron to run update_topology command

rohithasrk:
References #20

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -39,3 +39,7 @@ openwisp2_sentry:
 openwisp2_default_cert_validity: 1825
 openwisp2_default_ca_validity: 3650
 openwisp2_default_organization_id: 78f74fdd-67c6-4064-97e1-ce761da30745
+openwisp2_topology_update_frequency:
+    day: "*"
+    hour: "*"
+    minute: "*/5"
--- a/tasks/cron.yml
+++ b/tasks/cron.yml
@@ -1,0 +1,9 @@
+- name: Install topology update cron
+  when: openwisp2_network_topology
+  become: true
+  cron:
+    name: "Update toplogies"
+    day: "{{ openwisp2_topology_update_frequency.day }}"
+    hour: "{{ openwisp2_topology_update_frequency.hour }}"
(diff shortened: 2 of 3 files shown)


tasks/cron.yml
```diff
@@ -0,0 +1,8 @@
+- name: Update topologies cron
```
nemesisdesign: add:

```yaml
when: openwisp2_network_topology
```

otherwise the cron will be installed also when the module is not active.

defaults/main.yml
```diff
@@ -39,3 +39,7 @@ openwisp2_sentry:
 openwisp2_default_cert_validity: 1825
 openwisp2_default_ca_validity: 3650
 openwisp2_default_organization_id: 78f74fdd-67c6-4064-97e1-ce761da30745
+openwisp2_topology_update_frequency:
+    day: "*"
+    hour: 0
```
nemesisdesign: set this to `*` please

defaults/main.yml
```diff
@@ -39,3 +39,7 @@ openwisp2_sentry:
...
+    minute: 0
```
nemesisdesign: set this to `*/5` to make it run every 5 minutes

tasks/cron.yml
```diff
@@ -0,0 +1,8 @@
...
+  become: yes
```
nemesisdesign: change this to `true` for consistency with the rest of boolean values used in the playbook

rohithasrk: Do you mean become: true?

nemesisdesign: yes

defaults/main.yml
```diff
@@ -39,3 +39,7 @@ openwisp2_sentry:
 openwisp2_default_cert_validity: 1825
 openwisp2_default_ca_validity: 3650
 openwisp2_default_organization_id: 78f74fdd-67c6-4064-97e1-ce761da30745
+openwisp2_topology_update_frequency:
+    day: *
+    hour: 0
```
rohithasrk: @nemesisdesign This is outdated maybe. I've updated the PR.

Issue #22, opened 2017-08-15
[network-topology] Add a cron to run save_snapshot command


rohithasrk:
A cron to run save_snapshot command. This issue can be referenced after merging https://github.com/openwisp/openwisp-network-topology/pull/17.

nemesisdesign:
closed by #23

PR #23, opened 2017-08-15
[topology] Added cron to run save_snapshot command

rohithasrk:
References #22

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -44,3 +44,7 @@ openwisp2_topology_update_frequency:
     day: "*"
     hour: "*"
     minute: "*/5"
+openwisp2_topology_save_snapshot_frequency:
+    day: "*"
+    hour: "23"
+    minute: "30"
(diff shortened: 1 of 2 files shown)


defaults/main.yml
```diff
@@ -44,3 +44,7 @@ openwisp2_topology_update_frequency:
     day: "*"
     hour: "*"
     minute: "*/5"
+openwisp2_topology_save_snapshot_frequency:
+    day: "*/1"
```
nemesisdesign: every day can be shortened to simply `*`

defaults/main.yml
```diff
@@ -44,3 +44,7 @@ openwisp2_topology_update_frequency:
...
+    hour: "0"
+    minute: "0"
```
nemesisdesign: let's start this process at 23:30 by default, that way we will capture the state of the network during the last hours of the day

nemesisdesign:
@rohithasrk could you rebase your branch on the current master branch so the final merge with rebase strategy will result cleaner?

rohithasrk:
Check now @nemesisdesign

PR #41, opened 2017-11-18
Added uid option in uwsgi.ini.j2

xamfy:
fixes #41

### DIFF
--- a/templates/uwsgi.ini.j2
+++ b/templates/uwsgi.ini.j2
@@ -1,5 +1,6 @@
 [uwsgi]
 chdir={{ openwisp2_path }}
+uid=www-data
 module=openwisp2.wsgi:application
 master=True
 pidfile={{ openwisp2_path }}/uwsgi.pid


tasks/uwsgi.yml
```diff
@@ -0,0 +1,11 @@
...
+- file:
+  path: {{ openwisp2_path }}
+  owner:  www-data
+  mode: 0744
+
+- file:
+  path: {{ openwisp2_path }}/log
+  owner:  www-data
+  mode: 0744
+
```
nemesisdesign: add a new line at the end of the file (configure your editor to do this automatically)

xamfy:
the log file goes in openwisp2_path only 
image

tasks/uwsgi.yml
xamfy: okay, also need to make some changes and remove the second path

tasks/uwsgi.yml
```diff
@@ -0,0 +1,6 @@
+# set directory permissions for uswgi for logs and pid file
+- file:
```
nemesisdesign: I found out this is not necessary because is already being performed here:
https://github.com/openwisp/ansible-openwisp2/blob/master/tasks/django.yml#L13-L18

I tested your patch without this task and it works anyway, so better remove this file and the related include in directive in `tasks/main.yml`, the rest is fine, thank you @xamfy!

xamfy:
@nemesisdesign Done

nemesisdesign:
@xamfy thank you! 👍

PR #67, opened 2017-12-30
[config] add gzip config to nginx

AZtheAsian:


defaults/main.yml
```diff
@@ -52,3 +52,9 @@ openwisp2_topology_save_snapshot_frequency:
     day: "*"
     hour: "23"
     minute: "30"
+openwisp2_nginx_ssl_config:
+  gzip: "on"
```
nemesisdesign: use 4 space indentation as the rest of the file does

defaults/main.yml
```diff
@@ -52,3 +52,9 @@ openwisp2_topology_save_snapshot_frequency:
...
+  gzip_comp_level: "6"
+  gzip_proxied: "any"
+  gzip_min_length: "1000"
+  gzip_types: "text/plain text/html image/svg+xml application/json application/javascript text/xml text/css application/xml application/x-font-ttf font/opentype"
```
nemesisdesign: can you turn this long string into a list of items that is then reassembled in the nginx template?

Eg: in the template you can probably use `{{ ' '.join(gzip_types) }}`

templates/nginx.j2
```diff
@@ -27,6 +27,13 @@ server {
     add_header Strict-Transport-Security "max-age=31536000";
     add_header X-Content-Type-Options nosniff;
 
+    gzip {{ openwisp2_nginx_ssl_config.gzip }};
+    gzip_comp_level {{ openwisp2_nginx_ssl_config.gzip }};
+    gzip_proxied {{ openwisp2_nginx_ssl_config.gzip }};
+    gzip_min_length {{ openwisp2_nginx_ssl_config.gzip }};
+    gzip_proxied {{ openwisp2_nginx_ssl_config.gzip }};
+    gzip_types {{ openwisp2_nginx_ssl_config.gzip }};
```
nemesisdesign: all these lines refer to `gzip` while they should refer their respective correct key

AZtheAsian:
@nemesisdesign got it fixed - sorry about that.

It works though!

got it

defaults/main.yml
```diff
@@ -52,3 +52,9 @@ openwisp2_topology_save_snapshot_frequency:
     day: "*"
     hour: "23"
     minute: "30"
+openwisp2_nginx_ssl_config:
+    gzip: "on"
+    gzip_comp_level: "6"
+    gzip_proxied: "any"
+    gzip_min_length: "1000"
+    gzip_types: ["text/plain", "text/html", "image/svg+xml", "application/json", "application/javascript", "text/xml", "text/css", "application/xml", "application/x-font-ttf", "font/opentype"]
```
nemesisdesign: use the extended format, eg:

```yaml
gzip_types:
    - item1
    - item2
```

AZtheAsian:
@nemesisdesign like this?

templates/nginx.j2
```diff
@@ -27,6 +27,14 @@ server {
     add_header Strict-Transport-Security "max-age=31536000";
     add_header X-Content-Type-Options nosniff;
 
+    {% for gzip_variable in openwisp2_nginx_ssl_config %}
```
nemesisdesign: you could have named this `config` since this dict can be used also for non-gzip related variables now

PR #73, opened 2018-01-11
[docs] Improve VM installation guide

HizkiFW:
Notes:

- Using port forwarding to set up the VM is not a good solution. It's much cleaner to assign an IP address to it instead, using VirtualBox's Host-only Adapter.
- Debian ships with Python and OpenSSH Server


UtkarshVerma:
But Host-Only disconnects the VM from the internet making ansible unable to download packages on the VM

HizkiFW:
@UtkarshVerma That's why a NAT adapter is also added alongside the Host-only adapter. This allows the VM to connect to the internet while having a separate IP it could be accessed from. It's mentioned in the linked guide.

UtkarshVerma:
@HizkiFW I tried using it and found out that sometimes the VM OS doesn't add the adapter by itself. That complicates the process a bit so I used NAT instead. I was unable to set this up on my system.

nemesisdesign:
I'm merging. If there's any issue users will report it.

PR #88, opened 2018-09-01
[variables][Fixes #77] Spatialite Path Changes According to OS

### LINKED ISSUE #77: Ansible install fails in migrate on Debian Buster and Ubuntu 18
Installation in Debian Buster (aka testing) fails on the migrate step. Not entirely sure why.
Running in x86_64.

```
TASK [openwisp.openwisp2 : migrate] **********************************************************************************************************************
fatal: [openwisp.home.lan]: FAILED! => {"changed": false, "cmd": "./manage.py migrate --noinput", "msg": "\n:stderr: Traceback (most recent call last):\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/gis/db/backends/spatialite/base.py\", line 51, in get_new_connection\n    conn.load_extension(self.spatialite_lib)\nsqlite3.OperationalError: mod_spatialite: cannot open shared object file: No such file or directory\n\nThe above exception was the direct cause of the following exception:\n\nTraceback (most recent call last):\n  File \"./manage.py\", line 10, in <module>\n    execute_from_command_line(sys.argv)\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/core/management/__init__.py\", line 371, in execute_from_command_line\n    utility.execute()\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/core/management/__init__.py\", line 365, in execute\n    self.fetch_command(subcommand).run_from_argv(self.argv)\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/core/management/base.py\", line 288, in run_from_argv\n    self.execute(*args, **cmd_options)\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/core/management/base.py\", line 335, in execute\n    output = self.handle(*args, **options)\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/core/management/commands/migrate.py\", line 77, in handle\n    connection.prepare_database()\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/gis/db/backends/spatialite/base.py\", line 61, in prepare_database\n    with self.cursor() as cursor:\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/db/backends/base/base.py\", line 255, in cursor\n    return self._cursor()\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/db/backends/base/base.py\", line 232, in _cursor\n    self.ensure_connection()\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/db/backends/base/base.py\", line 216, in ensure_connection\n    self.connect()\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/db/backends/base/base.py\", line 194, in connect\n    self.connection = self.get_new_connection(conn_params)\n  File \"/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/gis/db/backends/spatialite/base.py\", line 55, in get_new_connection\n    ) from exc\ndjango.core.exceptions.ImproperlyConfigured: Unable to load the SpatiaLite library extension \"mod_spatialite\"\n", "path": "/opt/openwisp2/env/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "state": "absent", "syspath": ["/tmp/ansible_5L_TsX", "/tmp/ansible_5L_TsX/ansible_modlib.zip", "/tmp/ansible_5L_TsX/ansible_modlib.zip", "/usr/lib/python2.7", "/usr/lib/python2.7/plat-x86_64-linux-gnu", "/usr/lib/python2.7/lib-tk", "/usr/lib/python2.7/lib-old", "/usr/lib/python2.7/lib-dynload", "/usr/local/lib/python2.7/dist-packages", "/usr/lib/python2.7/dist-packages"]}

```

atb00ker:
- `openwisp2_spatialite_path` value changes according to `ansible_distribution` & `ansible_distribution_version`
- added `.travis.yml` & `.gitlab-ci.yml` jobs for `debian 10` and `ubuntu 18.04`

Fixes #77

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,6 +35,13 @@ Ubuntu1604:
 
   <<: *testing_script
 
+Ubuntu1804:
+  variables:
+    distro: 'ubuntu:18.04'
+    playbook: test.yml
+
+  <<: *testing_script
+
 Centos7:
   variables:
     distro: centos:7
@@ -56,9 +63,23 @@ Debian9:
 
   <<: *testing_script
 
-Ubuntu1604_NetworkTopology:
+Debian10:
   variables:
-    distro: 'ubuntu:16.04'
+    distro: debian:10
+    playbook: test.yml
+
+  <<: *testing_script
+
+Fedora27:
+  variables:
+    distro: fedora:27
+    playbook: test.yml
+
+  <<: *testing_script
+
+Ubuntu1804_NetworkTopology:
+  variables:
+    distro: 'ubuntu:18.04'
     playbook: test_networktopology.yml
 
   <<: *testing_script
--- a/README.md
+++ b/README.md
@@ -436,7 +436,9 @@ Below are listed all the variables you can customize (you may also want to take 
         port: ""
         options: {}
     # SPATIALITE_LIBRARY_PATH django setting
-    openwisp2_spatialite_path: "mod_spatialite"
+    # The role will attempt determining the right mod-spatialite path automatically
+    # But you can use this variable to customize the path or fix future arising issues
+    openwisp2_spatialite_path: null
     # customize other django settings:
     openwisp2_language_code: en-gb
     openwisp2_time_zone: UTC
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -32,15 +32,6 @@
   when: openwisp2_database.engine == "django.contrib.gis.db.backends.spatialite"
   ignore_errors: yes
   apt: name=libsqlite3-mod-spatialite state=latest
-  register: mod_spatialite_result
-
-- name: Set openwisp2_spatialite_path
-  set_fact:
-    openwisp2_spatialite_path: false
-  when: >
-    openwisp2_database.engine != "django.contrib.gis.db.backends.spatialite"
-    or (openwisp2_spatialite_path == 'mod_spatialite'
-        and mod_spatialite_result is failed)
 
 # fixes issue described in https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user
 - name: Install acl if acting as non-root user
--- a/tasks/variables-spatialite.yml
+++ b/tasks/variables-spatialite.yml
@@ -1,0 +1,22 @@
+- name: Set spatialite_path for Fedora
+  set_fact:
+    openwisp2_spatialite_path: "mod_spatialite"
+  when: ansible_distribution == 'Fedora'
+
+- name: Set spatialite_path (Ubuntu >= 18.04 or Debian >= 10)
+  set_fact:
+    openwisp2_spatialite_path: "mod_spatialite.so"
+  when: >
+    (ansible_distribution == 'Ubuntu' 
+    and ansible_distribution_version is version_compare('18.04', 'ge')) 
+    or (ansible_distribution == 'Debian' and 
+    ansible_distribution_version is version_compare('10', 'ge'))
+
+- name: Set spatialite_path (Ubuntu >= 16.04 or Debian >= 9)
+  set_fact:
+    openwisp2_spatialite_path: "mod_spatialite"
+  when: >
+    (ansible_distribution == 'Ubuntu'
+    and ansible_distribution_version is version_compare('16.04', 'ge'))
+    or (ansible_distribution == 'Debian'
+    and ansible_distribution_version is version_compare('9', 'ge'))
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,18 @@ env:
     playbook: test.yml
   - distro: 'ubuntu:16.04'
     playbook: test.yml
+  - distro: 'ubuntu:18.04'
+    playbook: test.yml
   - distro: debian:8
     playbook: test.yml
   - distro: debian:9
     playbook: test.yml
+  - distro: debian:10
+    playbook: test.yml
+  - distro: fedora:27
+    playbook: test.yml
 
-  - distro: 'ubuntu:16.04'
+  - distro: 'ubuntu:18.04'
     playbook: test_networktopology.yml
 
 script:
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -23,7 +23,7 @@ openwisp2_database:
     host: ""
     port: ""
     options: {}
-openwisp2_spatialite_path: "mod_spatialite"
+openwisp2_spatialite_path: null
 openwisp2_language_code: "en-gb"
 openwisp2_time_zone: "UTC"
 openwisp2_context: {}
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -5,6 +5,12 @@
 - import_tasks: variables.yml
   tags: [openwisp2, apt, yum, pip, django, supervisor, nginx, variables]
 
+- import_tasks: variables-spatialite.yml
+  tags: [openwisp2, apt, yum, pip, django, supervisor, nginx, variables]
+  when: >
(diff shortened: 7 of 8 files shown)


atb00ker:
Tests currently failing because docker files for `ubuntu18.04` and `debian10` do not exist.

tasks/variables.yml
```diff
@@ -29,3 +29,26 @@
...
+     and ansible_distribution_version is version_compare('10', 'lt'))
+
+- name: Set Spatialite Path (Ubuntu >= 18.4 or Debian >= 10)
+  set_fact:
+    openwisp2_spatialite_path: "mod_spatialite.so"
+  when: >
+    (ansible_distribution == 'Ubuntu' 
+    and ansible_distribution_version is version_compare('18.04', 'ge')) 
+    or (ansible_distribution == 'Debian' and 
+    ansible_distribution_version is version_compare('10', 'ge'))
```
nemesisdesign: `openwisp2_database.engine == "django.contrib.gis.db.backends.spatialite"`

nemesisdesign:
PS: rebase your branch on the current master, eg:

```
git checkout issues/77
git rebase master
```

README.md
```diff
@@ -435,8 +435,6 @@ Below are listed all the variables you can customize (you may also want to take
         host: ""
         port: ""
         options: {}
-    # SPATIALITE_LIBRARY_PATH django setting
-    openwisp2_spatialite_path: "mod_spatialite"
```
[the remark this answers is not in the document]
atb00ker: Since `variable.yml` runs everytime and the value set by the file takes precedence over value in `playbook.yml`; this variable will have no effect.
If we want `playbook.yml`'s values to have precedence, we'll need to take `openwisp2_spatialite_path` to a file with lower precedence, should i do that?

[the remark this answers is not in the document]
atb00ker: After, `openwisp2_spatialite_path: false` is added to `default/main.yml` and condition to not change value from `variables.yml` for RedHat system is added. This value from `playbook.yml` will take effect in RedHat systems, but this will be a little inconsistent because the value will still be obsolete for other OS.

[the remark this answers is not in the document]
nemesisdesign: you're right, so we could put `null` instead and execute `set_fact` only if `openwisp2_spatialite_path == null`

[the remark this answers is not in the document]
atb00ker: That should do the trick, thanks, i am on it! :smile:

hispanico:
I've build Ubuntu 18.04 docker image, but I'm not sure to include Debian 10 as well because is not the stable release.

nemesisdesign:
@hispanico in the worst case we can run tests on ubuntu 18 only..

tasks/apt.yml
```diff
@@ -37,10 +37,7 @@
 - name: Set openwisp2_spatialite_path
   set_fact:
     openwisp2_spatialite_path: false
-  when: >
-    openwisp2_database.engine != "django.contrib.gis.db.backends.spatialite"
-    or (openwisp2_spatialite_path == 'mod_spatialite'
-        and mod_spatialite_result is failed)
+  when: mod_spatialite_result is failed
```
atb00ker: Made all the changes except one:
I have not removed the following code yet:
```
  register: mod_spatialite_result

- name: Set openwisp2_spatialite_path
  set_fact:
    openwisp2_spatialite_path: false
  when: >
    openwisp2_database.engine != "django.contrib.gis.db.backends.spatialite"
    or (openwisp2_spatialite_path == 'mod_spatialite'
        and mod_spatialite_result is failed)
```
I am thinking since the code is still relevant in case user tries to install in older ubuntu / debian versions.
Example: Ubuntu
since `openwisp2_spatialite_path: "mod_spatialite"` is set for version less than 18.04, 
older versions like 12.04 will still fail and would require `openwisp2_spatialite_path: false` to be set to false.

However, i have modified the condition, since `openwisp2_database.engine != "django.contrib.gis.db.backends.spatialite"` doesn't need to changed again (as it was checked in the task above) and `openwisp2_spatialite_path` could have different values.

tasks/variables-spatialite.yml
```diff
@@ -0,0 +1,26 @@
+- name: Set spatialite_path for Fedora
+  set_fact:
+    openwisp2_spatialite_path: "mod_spatialite"
+  when: ansible_distribution == 'Fedora'
```
nemesisdesign: Fedora doesn't have mod-spatialite, I think you can remove this (`null` will be used on RedHat based systems).

tasks/variables-spatialite.yml
```diff
@@ -0,0 +1,26 @@
...
+    openwisp2_spatialite_path: "mod_spatialite.so"
+  when: >
+    (ansible_distribution == 'Ubuntu' 
+    and ansible_distribution_version is version_compare('18.04', 'ge')) 
+    or (ansible_distribution == 'Debian' and 
+    ansible_distribution_version is version_compare('10', 'ge'))
+
+- name:  Set spatialite_path for Other System (like RedHat & CentOS)
+  set_fact:
+    openwisp2_spatialite_path: false
```
nemesisdesign: this is wrong, it was present in `yum.yml` which was called only when the distribution is redhat based, now this would be always executed, invalidating the previous lines.

Simply remove this because it's not needed anymore.

tasks/variables-spatialite.yml
atb00ker: I ran it on my system again, it looks like in fedora-27, it requires `mod_spatialite`, it fails otherwise!
Should i try on some other fedora version as well?

atb00ker: I just checked with the version on anisble-galaxy, that's failing because of `mod_spatialite` too!
(On fedora 27)

atb00ker: As per my understanding right now, the value of `openwisp2_spatialite_path` is required according to the table below:

OS | Version |
--- | --- | 
Ubuntu:14.04 | false
Ubuntu:16.04 | mod_spatialite
Ubuntu:18.04 | mod_spatialite.so
Debian:8 | false
Debian:9 | mod_spatialite
Debian:10 | mod_spatialite.so
Fedora:27 | mod_spatialite

nemesisdesign: ok if you tested it on Fedora27 and it's required let's leave it

tasks/variables-spatialite.yml
nemesisdesign: I still wonder why is this here? Am I missing something?

.gitlab-ci.yml
```diff
@@ -56,9 +63,23 @@ Debian9:
...
   variables:
     distro: 'ubuntu:16.04'
     playbook: test_networktopology.yml
 
   <<: *testing_script
+
+Ubuntu1804_NetworkTopology:
+  variables:
+    distro: 'ubuntu:18.04'
+    playbook: test_networktopology.yml
```
nemesisdesign: one little request: could you simply replace `Ubuntu1604_NetworkTopology` with this block instead of adding a new one? The network topology module can be tested just once. We don't want the build time to increase too much.

atb00ker: Sure! On it! :smile:

atb00ker: Done! Added build for `fedora:27` as well.
Note: `fedora:28` not added because it will requires more changes,
I think i should create another issue for it! :smile:

nemesisdesign: @atb00ker yes please create a new issue for `fedora28`

PR #100, opened 2018-09-28
[ci] Avoiding duplication of basic test by a script #84

### LINKED ISSUE #84: [ci] Avoid duplication of basic running test
These lines seem duplicated:

- https://github.com/openwisp/ansible-openwisp2/blob/master/.travis.yml#L31
- https://github.com/openwisp/ansible-openwisp2/blob/master/.gitlab-ci.yml#L16-L19

Maybe we can extract those lines in a script that gets called by both CI systems

armudgal:
Added a script accessed by both travis and gitlab-ci to run
openwisp and test it

Closes #84

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,11 +15,7 @@
     - ${PWD}/tests/test.sh
 
     # Make sure OpenWISP is running.
-    - >
-      docker exec ${container_id} curl --insecure -s --head https://localhost/admin/login/?next=/admin/
-      | sed -n '1p' | grep -q '200'
-      && (echo 'Status code 200 test: pass' && exit 0)
-      || (echo 'Status code 200 test: fail'; docker exec ${container_id} sh -c 'tail -n 500 /opt/openwisp2/log/*.log' && exit 1)
+    - ./tests/runtests.sh
   dependencies: []
   after_script:
     - docker rm -f $container_id
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,11 +41,7 @@ script:
   - ${PWD}/tests/test.sh
 
   # Make sure OpenWISP is running.
-  - >
-    docker exec ${container_id} curl --insecure -s --head https://localhost/admin/login/?next=/admin/
-    | sed -n '1p' | grep -q '200'
-    && (echo 'Status code 200 test: pass' && exit 0)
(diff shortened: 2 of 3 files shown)


hispanico:
I think is better to put the script in some github gist (https://gist.github.com/) or gitlab snippet (https://gitlab.com/snippets/) to keep the ansible role folder cleaned. Then get the script in travis-ci and gitlab-ci with commands like following: 
```
# Download test shim.
  - wget -O ${PWD}/tests/test.sh https://gitlab.com/snippets/1751673/raw
  - chmod +x ${PWD}/tests/test.sh
```

.gitlab-ci.yml
```diff
@@ -15,11 +15,7 @@
     - ${PWD}/tests/test.sh
 
     # Make sure OpenWISP is running.
-    - >
-      docker exec ${container_id} curl --insecure -s --head https://localhost/admin/login/?next=/admin/
-      | sed -n '1p' | grep -q '200'
-      && (echo 'Status code 200 test: pass' && exit 0)
-      || (echo 'Status code 200 test: fail'; docker exec ${container_id} sh -c 'tail -n 500 /opt/openwisp2/log/*.log' && exit 1)
+    - ./scripts/run-openwisp.sh
```
nemesisdesign: `./scripts/run-openwisp.sh` is not helpful as a name and as a path. I'd simply use `./tests/runtests.sh` as it's at least similar with the `runtests.py` and `runtests` scripts we have in other repos.

hispanico:
> @hispanico since the code of the tests are directly related to this project, I don't see how extracting that logic into a separate repository would help us, I think it would instead make things more complicated.
> 
> The test logic should evolve along with the role so it makes sense to keep it here. I don't see any advantage in extracting it. What kind of advantage do you see? Keep in mind removing code from a repo/package is good only when the code is not useful or not related to the package.

@nemesisdesign that's make sense :D 
Some times my mind is to muche general purpose :D

nemesisdesign:
@hispanico no problem, I'll help you out in this kind of decisions, I banged my head on this wall enough times that I became able to give others good suggestions now

.gitlab-ci.yml
armudgal: HI, +1 to the comment, sorry have been busy lately :)

nemesisdesign:
Thanks @armudgal

PR #113, opened 2018-10-25
[apt + pip] Fix Ansible deprecation warnings #86

### LINKED ISSUE #86: Deprecation Warning in openwisp.openwisp2 set_fact task
TASK [openwisp.openwisp2 : set_fact] *******************************************
[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using
result|failed use result is failed. This feature will be removed in version
 2.9. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.

dwang:
The deprecation warnings referenced in the issue are not present anymore, but there were other warnings, which I fixed.

Ansible gives deprecation warnings for the "apt" and "pip" tasks. For example: 
```
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of
using a loop to supply multiple items and specifying `name: {{ item }}`, please use `name: ['sudo', 'software-
properties-common', 'build-essential', 'supervisor', 'nginx', 'openssl', 'libssl-dev', 'libffi-dev', 'python-dev',
'redis-server']` and remove the loop. This feature will be removed in version 2.11. Deprecation warnings can be
disabled by setting deprecation_warnings=False in ansible.cfg
```

Now, the log is free of any deprecation warnings.

Tests run using `tests/runtests.sh` pass.

Fixes #86

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -3,21 +3,22 @@
   changed_when: false
 
 - name: Install system packages
-  apt: name={{ item }} state=latest
-  notify: 
+  apt:
+    name:
+      - sudo
+      - software-properties-common
+      - build-essential
+      - supervisor
+      - nginx
+      - openssl
+      - libssl-dev
+      - libffi-dev
+      - python-dev
+      - redis-server
+    state: latest
+  notify:
     - reload systemd
     - start redis-server
-  with_items:
-    - sudo
-    - software-properties-common
-    - build-essential
-    - supervisor
-    - nginx
-    - openssl
-    - libssl-dev
-    - libffi-dev
-    - python-dev
-    - redis-server
 
 # On the newer versions of redis, by default redis
 # binds to localhost on ipv6 address which wouldn't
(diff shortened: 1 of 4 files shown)


tasks/apt.yml
```diff
@@ -3,8 +3,20 @@
   changed_when: false
 
 - name: Install system packages
-  apt: name={{ item }} state=latest
-  notify: 
+  apt:
+    name:
+        - sudo
+        - software-properties-common
```
nemesisdesign: looks like you're using 4 spaces indentation here (and in other sections below as well), while the rest of the playbook uses 2 spaces, could you make it consistent please?

dwang:
Sorry about that!
I'll work on it when I have time (probably in a few hours)

dwang:
Just amended the commit. Formatting should be better (hopefully). Please tell me if you have any other issues!

dwang:
Actually, I think that removing `with_items` will work. I will try that now.

dwang:
Travis tests are passing now! 🎉

hispanico:
> Since when is this new way of doing things available?

Ansible 2.5+:
> A list of package names, like foo, or package specifier with version, like foo=1.0. Name wildcards (fnmatch) like apt* and version wildcards like foo=1.0* are also supported.

tasks/apt.yml
```diff
@@ -28,19 +29,20 @@
   lineinfile:
     path: /etc/redis/redis.conf
     regexp: '^bind 127\.0\.0\.1 ::1'
-    line: 'bind 127.0.0.1'
+    line: "bind 127.0.0.1"
```
atb00ker: Why was this changed to `"`?

atb00ker:
@hispanico 
I am unable to reproduce https://github.com/openwisp/ansible-openwisp2/issues/86.
I tried different versions of ansible as well. Is that fixed?

tasks/apt.yml
dwang: Reverted and amended commit

Issue #93, opened 2018-09-08
Fails on Fedora 28


atb00ker:
Setup Fails on Fedora 28;
Fail log:
```
TASK [role_under_test : Install firewall packages] *****************************
failed: [localhost] (item=python-firewall) => {"changed": false, "item": "python-firewall", "msg": "No package python-firewall available."}
changed: [localhost] => (item=firewalld)
```

nemesisdesign:
Does it still fail? If yes @atb00ker can you create a GCI task for this please? And If not close please.

1871vinayak:
I'll take this issue 👍

PR #122, opened 2018-12-10
[qa] Fix Travis CI build errors #93

dwang:
Travis CI builds for Debian 9, Debian 10, and Fedora 28 have been fixed. 
- Python3 is set as the `ansible_python_interpreter` for Fedora 28 because it drops support for `python-firewall` which is for Python2. 
- Cron is also added to `apt.yml` to resolve errors in Debian 10. 
- Additionally, `cryptography` from pip is installed to fix an odd SSL error when installing `django-redis` on Debian 9.
- The `spatialite_path` is updated for Fedora 28

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,6 @@ env:
   - distro: 'ubuntu:18.04'
     playbook: test_networktopology.yml
 
-matrix:
-  allow_failures:
-    - env: distro=debian:10 playbook=test.yml
-    - env: distro=fedora:28 playbook=test.yml
-
 script:
   # Configure test script so we can run extra tests after playbook is run.
   - export container_id=$(date +%s)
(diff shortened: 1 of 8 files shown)


dwang:
@atb00ker @hispanico 
Finally, all the builds are passing after 30 pushes 🎉 
But to do so, I had to move the [common tests shim](https://gitlab.com/snippets/1751673/raw) to `runtests.sh` in order to set `ansible_python_interpreter` to use Python3 for Fedora 28. Is this okay? 

Also, is there any particular reason to store the shim separately in a GitLab Snippet? I did not understand that.

Thanks

atb00ker:
@dwang about `runtests.sh`,
The reason is that the file used in multiple places and it's not practical to maintain multiple copies of the file, hence, it's called from GitLab. :smile:

dwang:
@atb00ker 
Thanks for clarifying, where else is it used as I want to modify the script to set `ansible_python_interpreter`. I worry that it might break whatever that is using the same tests shim.

@unsuitable001 I think it's stored on GitLab because @hispanico is more active there and he maintains the tests shim. It shouldn't matter as the repo on GitLab is mirrored from GitHub and the tests shim is stored in a GitLab Snippet (similar to a GitHub Gist).

PR #128, opened 2019-01-05
[meta/main.yml] Added new platform #127

### LINKED ISSUE #127: [meta/main.yml] platforms supported is outdated
Platforms supported lists trusty, which doesn't work anymore.
Moreover, some new platforms are added, platform tag needs to be updated as per travis jobs.

Dhiraj240:
As per the .travis.yml build matrix some
new platforms were added and older one
were removed

Fixes #127

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -16,17 +16,15 @@ galaxy_info:
   platforms:
   - name: Debian
     versions:
-    - wheezy
     - jessie
     - stretch
+    - buster
   - name: Ubuntu
     versions:
-    - trusty
     - xenial
     - bionic
   - name: Fedora
     versions:
-    - 25
     - 27
     - 28
   - name: EL


Dhiraj240:
@nemesisdesign so what could be the probably change here. should i exclude the updation of debian in this ```main.yml```

meta/main.yml
```diff
@@ -21,15 +21,14 @@ galaxy_info:
...
-    - trusty
     - xenial
     - bionic
   - name: Fedora
     versions:
     - 25
     - 27
     - 28
-  - name: EL
+  - name: centos
```
nemesisdesign: I looked up and it seems `centos` is not used on ansible-galaxy (EL comprises Centos too), therefore please remove this change.

nemesisdesign: @Dhiraj240 can you update the PR please?

PR #134, opened 2019-02-22
[ci] resolve ansible-lint errors

dwang:
I've been seeing some odd errors on Ansible recently and some new `ansible-lint` warnings popped up since the last PR. Hopefully, this solves all of them and builds can pass again.

For example: [Travis CI](https://travis-ci.org/openwisp/ansible-openwisp2/jobs/496595884#L621)

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -1,6 +1,10 @@
 - name: Update APT package cache
   apt: update_cache=yes
   changed_when: false
+  retries: 5
+  delay: 10
+  register: result
+  until: result is success
 
 - name: Install system packages
   apt:
@@ -19,6 +23,11 @@
   notify:
     - reload systemd
     - start redis-server
+  ignore_errors: yes
+  retries: 5
+  delay: 10
+  register: result
+  until: result is success
 
 # On the newer versions of redis, by default redis
 # binds to localhost on ipv6 address which wouldn't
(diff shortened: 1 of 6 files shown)


tasks/apt.yml
```diff
@@ -1,6 +1,10 @@
 - name: Update APT package cache
```
atb00ker: @hispanico 
Someone recently informed me it's not a good practice to include `update_cache` in role, rather it's better to add it in playbook directly like:

```
  pre_tasks:
    - name: update apt cache
      apt:
        update_cache: yes
        cache_valid_time: 86400
```
Reason:
Updating the package cache is very time consuming (you can even get blocked if you do it too frequently). For that reason it doesn't belong in individual roles, it should be done only once at the start of the playbook.

Can you please comment on the same? Should we change accordingly?

Thanks

tasks/apt.yml
```diff
@@ -1,6 +1,10 @@
...
   apt: update_cache=yes
   changed_when: false
+  retries: 5
```
atb00ker: @dwang sorry, i am not really aware of what was causing the failure and how does this help, can you please share your research? 

Thanks.

tasks/apt.yml
dwang: I'm not familiar with this, but I have seen that updating the package cache is time consuming. The change sounds beneficial to me.

tasks/apt.yml
dwang: @atb00ker There was not a failure directly related to the playbook, but with `ansible-lint`. It was complaining that there was not a retry in the case that remote tasks failed.
For example:
```
[405] Remote package tasks should have a retry
/home/travis/.ansible/roles/role_under_test/tasks/apt.yml:1
Task/Handler: Update APT package cache
```

With the PR, I fixed some more `ansible-lint` errors. I'm not sure why the lint errors weren't present before, but it was causing builds for master and other PRs to fail.

atb00ker: Alright thanks. :+1:

tasks/apt.yml
nemesisdesign: I prefer if we leave this how it is now, I'd fix this role and then start focusing the devops technical effort the dockerization project

tasks/apt.yml
nemesisdesign: good call :+1:

PR #137, opened 2019-04-04
[pip] Fix condition

okraits:
ansible 2.7.5 didn't accept https://github.com/openwisp/ansible-openwisp2/blob/f807dce65419a37d5111e324ce1b91163120e3f3/tasks/pip.yml#L15 here. So I improved that line. And it makes sense to use to `!=False` because those `openwisp2_*_pip` always exist through https://github.com/openwisp/ansible-openwisp2/blob/master/defaults/main.yml.
Additionally I improved the code formatting.

### DIFF
--- a/tasks/pip.yml
+++ b/tasks/pip.yml
@@ -12,7 +12,7 @@
     - "{{ openwisp2_django_x509_pip }}"
     - "{{ openwisp2_django_loci_pip }}"
     - "{{ openwisp2_netjsonconfig_pip }}"
-  when: item
+  when: item != False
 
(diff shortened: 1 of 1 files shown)


okraits:
@nemesisdesign How do we continue with this one?

I agree that `!=False` shouldn't be needed but without that the playbook fails here with Ansible 2.7.5 if I use one of those `openwisp2_*_pip` variables in the playbook. Can you reproduce that?

nemesisdesign:
Never had this issue, using ansible 2.9.3 right now. I'm closing this.

PR #138, opened 2019-04-15
[pip] Add openwisp2_utils_pip variable

okraits:
This one was missing in the ansible role.


okraits:
@nemesisdesign I force-pushed without a change to trigger a rebuild but the centos 7 build keeps failing. Is that a known problem? How can I merge my change nevertheless?

atb00ker:
The default python version on travis-ci.org is now changed to python3 and I think this error is related to that.

okraits:
@nemesisdesign If you could please re-add your review, then I can merge this.

PR #140, opened 2019-04-21
[pip] Updated when syntax

nemesisdesign:
@okraits I bumped into the same issue you had.


okraits:
@nemesisdesign Great that you found a way to solve it! :+1:

PR #144, opened 2019-10-05
[pip + tests] Fix Ansible deprecation warning and minor issues

### LINKED ISSUE #126: deprecation warning in Install extra python
- TASK [openwisp.openwisp2 : Install extra python packages] ******************************************************************************
[DEPRECATION WARNING]: Invoking "pip" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply 
multiple items and specifying `name: "{{ item }}"`, please use `name: '{{ openwisp2_extra_python_packages }}'` and remove the loop. 
This feature will be removed in version 2.11. Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.

dwang:
- The deprecation warning from Ansible in the `Install extra python packages` task is fixed
- Package names are moved to the same line as `name:` if there is only one package installed
- Single quotes are replaced with double quotes to maintain consistency with the rest of the code
- Changed `Lauching` to `Launching` in `tests/runtests.sh`
- The removal of the Docker container is moved to the end of the `tests/runtests.sh` script
```
Removing Docker container...
1570316059
Lauching OpenWISP tests
Error: No such container: 1570316059
Status code 200 test: fail
Error: No such container: 1570316059
```
If this is not desired, it can exit before running the tests instead:
```bash
# Remove the Docker container (if configured).
if [ "$cleanup" = true ]; then
  printf "Removing Docker container...\n"
  docker rm -f $container_id
  exit 0
fi
```

Fixes #126

### DIFF
--- a/tasks/pip.yml
+++ b/tasks/pip.yml
@@ -16,7 +16,8 @@
   when: item | bool
 
 - name: Add network_topology to custom package list if set and enabled
-  set_fact: openwisp2_python_packages:"{{ openwisp2_python_packages + [item] }}"
+  set_fact:
+    openwisp2_python_packages: "{{ openwisp2_python_packages + [item] }}"
   with_items:
     - "{{ openwisp2_django_netjsongraph_pip }}"
     - "{{ openwisp2_network_topology_pip }}"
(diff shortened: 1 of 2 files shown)


nemesisdesign:
Debian 8 is failing with this error: "ImportError: Twisted on Python 3 requires Python 3.5 or later".
Shall we remove it?

atb00ker:
A lot of dependencies of Python 3.5 below have lost / are about to lose support.

PR #155, opened 2019-12-19
[bug] Reload systemd only if host have it #131

### LINKED ISSUE #131: [bug] run handler to reload systemd only when system is installed
`reload systemd` handler will fail the role when system is not running on systemd. 
Running that handler only when `systemd` is installed should solve the problem.

oltarzewskik:
Reload systemd only if host is running on it.

Tested on devuan linux (debian without systemd)
Fixes #131

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -4,7 +4,7 @@
 
 - name: reload systemd
   systemd: daemon_reload=yes
-  when: ansible_distribution_release in ['xenial', 'bionic']
+  when: ansible_distribution_release in ['xenial', 'bionic'] and ansible_service_mgr == 'systemd'
 
 - name: reload supervisor
   command: supervisorctl reload


oltarzewskik:
@nemesisdesign rebased :+1:

dwang:
I think that the check `ansible_distribution_release in ['xenial', 'bionic']` can be removed since it is redundant now right?

atb00ker:
> I think that the check ansible_distribution_release in ['xenial', 'bionic'] can be removed since it is redundant now right?

I can't recall right now why it was added, but was it related to this issue or something else?

oltarzewskik:
@atb00ker @dwang
> I can't recall right now why it was added, but was it related to this issue or something else?

That change was added in this commit: https://github.com/openwisp/ansible-openwisp2/commit/2fcc3412ae7248e320e7fbf33a5a8deb634e3183#diff-3d2b7487e8f37272aeb6f25599f2734b

PR #167, opened 2020-02-11
[README] Add basic system requirements

### LINKED ISSUE #102: [docs] Document software and hardware requirements
Questions like the following are recurrent:

- are there any packages I should pre-install prior to running the Ansible playbook?
- Are there any configuration steps I should perform on the server prior to running the playbook to ensure a clean install?
- what are the Hardware recommendations?

nickberry17:
Signed-off-by: Nicholas Smith <mips171@icloud.com>

Closes #102.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,41 @@ to know more about what happens under the hood in **OpenWISP 2**.
 
 For more information see also [Applying the Unix Philosophy to Django projects: a report from the real world](https://www.slideshare.net/FedericoCapoano/applying-the-unix-philosophy-to-django-projects-a-report-from-the-real-world).
 
+System Requirements
+===================
+
+The following specifications will run a new, *empty* instance of OpenWISP.
+Please ensure you account for the amount of disk space your use case will require, e.g. allocate
+enough space for users to upload floor plan images.
+
+Hardware (Minimum, excluding uploaded data)
+-------------------------------------------
+- 1GB Memory
+- 2GB free disk space
+- Single core processor
+
+Hardware (Recommended)
+----------------------
+- 2GB Memory
+- Disk space - depends on the projected size of your database and uploaded photo images
+- Dual core processor
+
+Software
+--------
+Generally a fresh installation of one of the supported operating systems is sufficient; no pre-configuration required.  The
+Ansible Playbook will install and configure all dependencies and leave you with a running OpenWISP installation.
+
+Tips: Ensure the hostname of your target machine matches what is in your Ansible configuration file.  Also, please ensure
+that Ansible can access your target machine by SSH, be it either with a key or password.  For more information see the
+[Ansible Getting Started Documentation](https://docs.ansible.com/ansible/latest/user_guide/intro_getting_started.html).
+
+Supported Operating Systems
+---------------------------
+- Ubuntu and Ubuntu Server 16.04
+- Ubuntu and Ubuntu Server 18.04
+- Debian 9
+- Debian 10
+
 Usage (tutorial)
 ================
 


README.md
```diff
@@ -34,6 +34,33 @@ to know more about what happens under the hood in **OpenWISP 2**.
 
 For more information see also [Applying the Unix Philosophy to Django projects: a report from the real world](https://www.slideshare.net/FedericoCapoano/applying-the-unix-philosophy-to-django-projects-a-report-from-the-real-world).
 
+System Requirements
+================
```
atb00ker: Can you please change it to same length as title:

```
System Requirements
===================
```

README.md
```diff
@@ -34,6 +34,33 @@ to know more about what happens under the hood in **OpenWISP 2**.
...
+
+The following specifications will run a new, empty instance of OpenWISP.  
+Please ensure you account for the amount of disk space your use case will require.
+
+Hardware
+---------------
```
atb00ker:
```
Hardware (Minimum)
------------------
```

README.md
```diff
@@ -34,6 +34,33 @@ to know more about what happens under the hood in **OpenWISP 2**.
...
+- 1GB Memory
+- 1GB disk space (excluding your database)
+- Single core processor
+
+Software
```
atb00ker:
```
Software
--------
```

README.md
```diff
@@ -34,6 +34,33 @@ to know more about what happens under the hood in **OpenWISP 2**.
...
+System Requirements
+================
+
+The following specifications will run a new, empty instance of OpenWISP.  
+Please ensure you account for the amount of disk space your use case will require.
+
+Hardware
+---------------
+- 1GB Memory
+- 1GB disk space (excluding your database)
```
okraits: 1 GB of disk space hardly is enough for the OS alone. I would suggest 5 GB of disk space. This should be enough for many usecases and provide enough space for experimentation.
If we want to specify the minimum disk space needed, I would suggest 2 or 3 GB (I don't know exactly how much disk space an installation with this ansible playbook currently uses).

@atb00ker What do you think?

nickberry17: I have a 50GB digital ocean droplet, running `df` shows roughly 48GB remaining with a running OpenWISP installation. It has 2GB RAM, running `free` shows about 1500MB free.  I agree bumping up to some recommended requirements would be reasonable.

atb00ker: ^ those numbers look good, thanks for sharing! 
I assumed that the 1GB is excluding the OS! Maybe we can explicitly say (1 GB free space)
Actually, the instance itself needs only a couple of 100 MBs, the main space requirements are for database and user uploaded data like floor plan images! :smile:

README.md
```diff
@@ -34,6 +34,41 @@ to know more about what happens under the hood in **OpenWISP 2**.
...
+===================
+
+The following specifications will run a new, *empty* instance of OpenWISP.
+Please ensure you account for the amount of disk space your use case will require, e.g. allocate
+enough space for users to upload floor plan images.
+
+Hardware (Minimum, excluding uploaded data)
+-------------------------------------------
+- 1GB Memory
+- 300MB free disk space
```
okraits: Is 300 MB of disk space really enough to be able to install OpenWISP with the ansible playbook? If the installation gets updated by running the playbook again, additional disk space might be used by package caches.
I just want to avoid that users run into trouble easily. And a couple of GB of disk space is not much these days.

README.md
```diff
@@ -34,6 +34,41 @@ to know more about what happens under the hood in **OpenWISP 2**.
...
+- 2GB Memory
+- Disk space - depends on the projected size of your database and uploaded photo images
+- Dual core processor
+
+Software
+--------
+Generally a fresh installation of one of the supported operating systems is sufficient; no pre-configuration required.  The
+Ansible Playbook will install and configure all dependencies and leave you with a running OpenWISP installation.
+
+Tips: Ensure the hostname of your device matches what is in your Ansible configuration file.  Also, please ensure
```
okraits: I would use a different word instead of `device` to avoid confusion with devices as e.g. routers. What about `target machine`?

nickberry17: good call

README.md
nickberry17: Let's just say 2GB free space to account for any future deps?

PR #183, opened 2020-05-06
[pip] Fixed install of custom packages

nemesisdesign:


tasks/pip.yml
```diff
@@ -107,32 +107,33 @@
...
+- name: Install custom OpenWISP 2 Python packages
+  when: openwisp2_python_packages
   pip:
-    name: "{{ openwisp2_extra_python_packages }}"
+    name: "{{ openwisp2_python_packages }}"
     state: latest
     virtualenv: "{{ virtualenv_path }}"
     virtualenv_python: "{{ openwisp2_python }}"
     virtualenv_site_packages: yes
+    extra_args: "--no-deps"
```
pandafy, line 131: I wonder if we create a `requirements.txt` of custom packages in `/tmp` of remote machine and use it, pip might resolve proper versions of dependencies automatically. 

@nemesisdesign @atb00ker

nemesisdesign: I believe that if we make sure all the dependencies of the dev versions are specified correctly using [PEP 508](https://www.python.org/dev/peps/pep-0508/) we can avoid specifying `--no-deps`.

atb00ker: If that will work, nice, but we will have to test it, this can also break the playbook in corner cases! :smile:

> dev versions are specified correctly 

But if the user installs their own custom module where deps are not that well managed, they'll face problem.

PR #188, opened 2020-06-14
[gsoc2020] Merged Modules remove django-netjsongraph

atb00ker:
Blocked by `openwisp-network-topology release` after https://github.com/openwisp/openwisp-network-topology/issues/66

### DIFF
--- a/tasks/pip.yml
+++ b/tasks/pip.yml
@@ -27,15 +27,14 @@
     - "{{ openwisp2_django_x509_pip }}"
     - "{{ openwisp2_django_loci_pip }}"
     - "{{ openwisp2_netjsonconfig_pip }}"
-  when: item is defined and item
+  when: item is defined and item is string
 
 - name: Add network_topology to custom package list if set and enabled
   set_fact:
     openwisp2_python_packages: "{{ openwisp2_python_packages + [item] }}"
   with_items:
-    - "{{ openwisp2_django_netjsongraph_pip }}"
     - "{{ openwisp2_network_topology_pip }}"
-  when: item and openwisp2_network_topology
+  when: item is defined and item is string and openwisp2_network_topology
 
 - name: Update pip & related tools
   pip:
@@ -124,7 +123,7 @@
   notify: reload supervisor
 
 - name: Install custom OpenWISP 2 Python packages
-  when: openwisp2_python_packages
+  when: openwisp2_python_packages|length > 0
   pip:
     name: "{{ openwisp2_python_packages }}"
     state: latest
--- a/README.md
+++ b/README.md
@@ -542,7 +542,7 @@ Below are listed all the variables you can customize (you may also want to take 
     openwisp2_controller_version: "0.4"
     # optional openwisp2 modules
     openwisp2_network_topology: false
-    openwisp2_network_topology_version: "0.2"
+    openwisp2_network_topology_version: "0.4"
     # you may replace the values of these variables with any URL
     # supported by pip (the python package installer)
     # use these to install forks, branches or development versions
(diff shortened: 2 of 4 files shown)


tasks/pip.yml
```diff
@@ -124,14 +122,13 @@
...
-  when: openwisp2_python_packages
+  when: openwisp2_python_packages|length > 0
   pip:
     name: "{{ openwisp2_python_packages }}"
-    state: latest
+    state: forcereinstall
     virtualenv: "{{ virtualenv_path }}"
     virtualenv_python: "{{ openwisp2_python }}"
     virtualenv_site_packages: yes
-    extra_args: "--no-deps"
```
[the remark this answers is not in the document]
nemesisdesign: when using custom targets (eg: the master branches) I found that the system would download and install the dependencies from pypi, overriding the custom dependencies just downloaded.

Eg:

- openwisp-utils 0.6.0 alpha is installed from github
- installing openwisp-controller master from github downloads and installs openwisp-utils 0.5.0 stable, hence losing openwisp-utils 0.6.0. I haven't found a solution to this yet, but if you do (it's quite likely it can be done since I didn't have a lot of time to dedicate to this) please let me know and maybe we can solve another painful point here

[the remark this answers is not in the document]
atb00ker: I am not able to replicate it,
I did:
```
openwisp2_utils_pip: https://github.com/atb00ker/openwisp-utils/tarball/master
openwisp2_controller_pip: https://github.com/openwisp/openwisp-controller/tarball/master
```

Where I randomly put util's version as `0.9.0` in my fork's master branch and I am seeing `0.9.0` only.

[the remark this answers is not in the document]
nemesisdesign: ok, let's remove it, I'll try an alternative and cleaner solution

tasks/pip.yml
```diff
@@ -113,7 +112,7 @@
 - name: Install extra python packages
   pip:
     name: "{{ openwisp2_extra_python_packages }}"
-    state: latest
+    state: forcereinstall
```
nemesisdesign, line 115: why is this needed?

tasks/pip.yml
```diff
@@ -122,21 +121,24 @@
   register: result
   until: result is success
   notify: reload supervisor
+  tags:
+    - molecule-idempotence-notest
```
nemesisdesign, line 125: why is this needed?

tasks/pip.yml
```diff
@@ -122,21 +121,24 @@
...
     virtualenv_python: "{{ openwisp2_python }}"
     virtualenv_site_packages: yes
-    extra_args: "--no-deps"
   notify: reload supervisor
   retries: 5
   delay: 10
   register: result
   until: result is success
+  tags:
+    - molecule-idempotence-notest
```
nemesisdesign, line 141: why is this needed?

tasks/pip.yml
atb00ker: Before `openwisp-network-topology` was released, I was getting an error swapper isn't installed because of `--no-deps` didn't allow dependencies to be installed. Hence, I removed `--no-deps`, then I realized that I was trying to install the same version of network-topology, but that wouldn't happen because it is already in `latest` (as version had not changed); I realized that a user may find themselves in the same problem trying to install their own version, so I changed it to `forceinstall`, which causes idempotence to fail, so I had to add `molecule-idempotence-notest`.

But after the release I am not able to reliably reproduce this problem with openwisp-utils. Hence, I am undoing all these changes. It's not specifically related to this PR and will stall this work needlessly.

But I think we should open a separate issue for this `--no-deps` problem.
Please let me know what do you think about it. :smile:

nemesisdesign: @atb00ker yes ok, let's open another issue for this :+1:

PR #190, opened 2020-06-18
[os] Added support for ubuntu 20

nemesisdesign:
- switched to use /usr/bin/python3 by default
- python-virtualenv > python3-virtualenv

Closes #189

PR #207, opened 2020-09-16
[feat] Added support for Ubuntu 20.04 (#189)

### LINKED ISSUE #189: [OS] Add support for Ubuntu 20 LTS
Add support for Ubuntu 20 LTS.

PabloCastellano:
(this PR integrates the work done in #190)

- switched to use /usr/bin/python3 by default
- python-virtualenv > python3-virtualenv
- build in travis using ubuntu 20.04
- add ubuntu 20.04 (focal) to meta info
- handle redis 4/5 and dependencies
- removed step to install python-wheel (installed by pip later)
- simplified python3-virtualenv installation
- pin channels_redis to 2.4 for redis 4 compatibility

Closes #189

### DIFF
--- a/files/generate_django_secret_key.py
+++ b/files/generate_django_secret_key.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """
 Pseudo-random django secret key generator
 """
--- a/files/generate_django_secret_key.py
+++ b/files/generate_django_secret_key.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
 Pseudo-random django secret key generator
 """
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
     - ROLE_NAME: openwisp2
   jobs:
     - MOLECULE_DISTRO: ubuntu1804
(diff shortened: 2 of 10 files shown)


files/generate_django_secret_key.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
nemesisdesign, line 1: are you sure this is needed? The virtualenv should know the python version without the need of introducing this change

nemesisdesign (commented):
Looks good, only one minor doubt below.

KapilBansal320:
@PabloCastellano the type should be feature instead of feat

nemesisdesign:
> @PabloCastellano the type should be feature instead of feat

More info here: https://github.com/openwisp/openwisp-utils/issues/110#issue-620578722

nemesisdesign (approved):
Fixed the minor issue https://github.com/openwisp/ansible-openwisp2/pull/207/commits/26654f0ec512f868599fd926a0f721b644058414.

Thanks!

files/generate_django_secret_key.py
PabloCastellano: Not sure now, but I probably made this change and then forgot to remove it :grin:

Issue #177, opened 2020-04-18
[bug] Install fail on debian 10


atb00ker:
A user reported on Gitter channel that installing on debian 10 cases an issue:
```
TASK [openwisp.openwisp2 : Remove jsonfield2] **
FAILED - RETRYING: Remove jsonfield2 (2 retries left).
FAILED - RETRYING: Remove jsonfield2 (1 retries left).
fatal: [(mydomain)]: FAILED! => {"attempts": 2, "changed": false, "msg": "Failed to find required executable virtualenv in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}

RUNNING HANDLER [Stouts.postfix : postfix restart] *

RUNNING HANDLER [openwisp.openwisp2 : reload systemd] **

RUNNING HANDLER [openwisp.openwisp2 : start redis] *

PLAY RECAP *
(mydomain) : ok=21 changed=11 unreachable=0 failed=1 skipped=25 rescued=0 ignored=0
```

Need to verify and fix it. :smile:

pniaps:
The same error with a fresh install of centos 7 in a VPS

```
TASK [openwisp.openwisp2 : Remove jsonfield2] **************************************************************************************************
FAILED - RETRYING: Remove jsonfield2 (2 retries left).
FAILED - RETRYING: Remove jsonfield2 (1 retries left).
fatal: [(mydomain)]: FAILED! => {"attempts": 2, "changed": false, "msg": "Failed to find required executable virtualenv in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin"}

RUNNING HANDLER [Stouts.postfix : postfix restart] *********************************************************************************************

PLAY RECAP *************************************************************************************************************************************
(mydomain)            : ok=9    changed=4    unreachable=0    failed=1    skipped=37   rescued=0    ignored=0   

```

pniaps:
My fault, CentOS is not supported.

nemesisdesign:
@pniaps are you getting this problem elsewhere? If yes can you paste the contents of `/etc/issue` please?

pniaps:
No, it worked on Ubuntu 18.04

nemesisdesign:
I'm glad it worked @pniaps, it should also work on ubuntu 20 if you use this branch: https://github.com/openwisp/ansible-openwisp2/tree/ubuntu20

miff2000:
[...]

PR #211, opened 2020-10-03
Move openwisp2_path directory creation to happen earlier

miff2000:
This should fix the missing virtualenv path error reported in #177

### DIFF
--- a/tasks/django.yml
+++ b/tasks/django.yml
@@ -1,9 +1,3 @@
-- name: create {{ openwisp2_path }}
-  file:
-    path: "{{ openwisp2_path }}"
-    state: directory
-    group: "{{ www_group }}"
-    mode: 0775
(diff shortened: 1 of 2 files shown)


atb00ker (commented):
I see.
This error was actually never faced by me, it was reported by a user.

Looking at the solution, it seems to make sense to move it and this should be the solution, However, I'll defer for another review on this one! :smile:

P.S: @miff2000 could you please rebase to master?

miff2000:
> P.S: @miff2000 could you please rebase to master?
That's been done :)

I've amended the commit and force pushed as the wrong email address was in it. I hope that won't affect you

nemesisdesign:
This change is unnecessary. Please try to focus your efforts on needed and useful changes.

PR #213, opened 2020-10-16
[fix] install molecule using docker driver and updated README

hispanico:
Fix molecule v3.1.1 regression: {'driver': [{'name': ['unallowed value docker']}]}


nemesisdesign (approved):
Thank you very much @hispanico!

PR #220, opened 2020-11-13
[feature] Added automatic generation of default SSH credentials #200

### LINKED ISSUE #200: [feature] Automatically create default SSH credentials
Feature needed to make push updates and firmware-upgrader work out of the box #199 

If there are no other credentials generated in the system:

- Automatically generate an SSH private key / public key pair
- Create a credentials object with root and the private key created, flag it as `auto_add=True`
- Look if there is any template flagged as default which contains /etc/dropbear/authorized_keys, if there is, append the public key to it, if there isn't, create one and flag it as default

pandafy:
A new SSH keypair will be generated if there does not exist any SSH
key pair on the machine.
The SSH key pair is used to create default access credential which
is used through a default template.
An existing template is updated if it contains configuration to
update the authorized_keys on devices.

Fixes #200

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -65,6 +65,7 @@
       - libproj-dev
       - libgeos-dev
       - libspatialite-dev
+      - openssh-client
   retries: 5
   delay: 10
   register: result
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -20,6 +20,8 @@
       - python3
       - virtualenv
       - cron
+      # needed to generate SSH key for push updates
+      - openssh-client
   ignore_errors: yes
   retries: 5
   delay: 10
@@ -65,7 +67,6 @@
       - libproj-dev
       - libgeos-dev
       - libspatialite-dev
-      - openssh-client
   retries: 5
   delay: 10
   register: result
--- a/tasks/ssh.yml
+++ b/tasks/ssh.yml
@@ -1,0 +1,19 @@
+- name: Create default SSH key pair
+  become: true
+  user:
+    name: root
+    generate_ssh_key: yes
+    ssh_key_bits: 4096
+    ssh_key_file: .ssh/id_rsa
+
+- name: Get default private SSH key
+  become: true
+  command: cat /root/.ssh/id_rsa
+  register: default_private_ssh_key
+  changed_when: false
+
+- name: Get default public SSH key
+  become: true
+  command: cat /root/.ssh/id_rsa.pub
+  register: default_public_ssh_key
+  changed_when: false
--- a/templates/load_initial_data.py
+++ b/templates/load_initial_data.py
@@ -3,6 +3,8 @@
 - Creates the admin user when openwisp2 is installed
 - Additionally creates the default organization if no organization is present
 - Modifies default Site object
+- Adds a default access credential and updates or creates default
+  template to use the same
 """
 import os
 import django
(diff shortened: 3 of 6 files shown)


tasks/ssh.yml
```diff
@@ -0,0 +1,6 @@
+- name: Create default SSH key pair
+  user:
+    name: "{{ www_user }}"
+    generate_ssh_key: yes
+    ssh_key_bits: 2048
```
nemesisdesign, line 5: let's use `4096` (strong)

templates/load_initial_data.py
```diff
@@ -29,3 +36,50 @@
...
+
+# Read SSH key pair
+with open('/home/{{www_user}}/.ssh/id_rsa', mode='r') as file:
+    ssh_private_key  = file.read()
+with open('/home/{{www_user}}/.ssh/id_rsa.pub', mode='r') as file:
+    ssh_pub_key  = file.read()
+
+# Create a default credentials object
+try:
+    Credentials.objects.create(
```
nemesisdesign, line 48: I think it's safe to create it only if no other credentials exist, the `IntegrityError` is a bit undesired since it generates noise in the postgres logs

templates/load_initial_data.py
```diff
@@ -29,3 +36,50 @@
...
+                },
+            ]
+        },
+    )
+    print('created Default Credentials Template')
+else:
+    for template_obj in queryset.iterator():
+        for config_file in template_obj.config['files']:
+            if config_file['path'] == '/etc/dropbear/authorized_keys':
+                if ssh_pub_key not in config_file['contents']:
```
nemesisdesign, line 82: I think these 2 ifs can be merged in one if statement

tasks/ssh.yml
```diff
@@ -0,0 +1,6 @@
+- name: Create default SSH key pair
+  user:
+    name: "{{ www_user }}"
```
nemesisdesign, line 3: I think we should use root and make sure the generated files have strict permissions (eg: 0600) so they're readable only by root. Using www-data means that in case of a security breach in the web application, if an attacker gains access to part of the filesystem and is able to read the home directory, it will be able to read the private key, which can have potential very negative consequences

nemesisdesign (changes requested):
Great progress @pandafy, see my comments below.

Try also to add a mention of this in the README (right where you added the suggestion to edit the site in the other PR).

pandafy:
@nemesisdesign can you please help me find out why idempotence tests are failing? I see nothing in `load_initial_data_result.stdout` while debugging.

tasks/ssh.yml
pandafy: The permissions are set automatically by the `user` module. 

In the update implementation, I have created separate task for reading generated keys. I thought running a script with escalated privileges was not a good idea.

nemesisdesign (commented):
This looks good, I need to test it.

tasks/apt.yml
```diff
@@ -65,6 +65,7 @@
       - libproj-dev
       - libgeos-dev
       - libspatialite-dev
+      - openssh-client
```
nemesisdesign, line 68: this should be added in the system dependencies, because if spatialite is not being used (one could use PostgreSQL or MySQL)

templates/load_initial_data.py
```diff
@@ -29,3 +35,48 @@
...
+                    'path': '/etc/dropbear/authorized_keys',
+                    'mode': '0644',
+                    'contents': ssh_pub_key,
+                },
+            ]
+        },
+    )
+    print('created Default Credentials Template')
+else:
+    for template_obj in queryset.iterator():
```
nemesisdesign, line 74: great choice to use `iterator()` here :+1:

PR #215, opened 2020-11-02
[feature] Added support for openwisp_notifications

pandafy:
- [x] Added support for openwisp_notifications integration in openwisp_controller
 - [x] Updated docs to ask user to update Site object
 - [x] Added message to ask user to update Site object when role completes execution
 - [x] Updated tests to run integration tests for notifications

Fixes #214 

--------

Blocked by:
 - [x] https://github.com/openwisp/openwisp-controller/pull/316 -  Refactoring of routes and URLs of openwisp_notifications
 - [x] https://github.com/openwisp/openwisp-notifications/pull/143 - Fixed stalling of migrate command when celery broker is unreachable
 - [x] https://github.com/openwisp/openwisp-notifications/pull/152 - Release openwisp-notifications==0.3.0
 - [x] Release openwisp-controller==0.8
------

I have used this playbook for manually testing this patch on Ubuntu 20.04

```
- hosts: openwisp2
  become: "{{ become | default('yes') }}"
  roles:
    - openwisp.openwisp2
  vars:
    openwisp2_controller_pip: https://github.com/pandafy/openwisp-controller/tarball/fix-notification-settings
    openwisp2_extra_python_packages:
      - django-flat-json-widget
      - https://github.com/openwisp/openwisp-notifications/tarball/master
    openwisp2_default_from_email: "openwisp2@openwisp2.mydomain.com"
```

PR #206, opened 2020-09-05
[deps] Added support for openwisp-controller 0.8.0

codesankalp:
`flat_json_widget` is now added to INSTALLED_APPS in the settings.py

Closes #197

------------------

**Blockers**

- [x] https://github.com/openwisp/openwisp-controller/pull/330

PR #223, opened 2020-11-21
{WIP}[feature] Added openwisp_radius installation

atb00ker:
Not even tested, my VM messed up today, I'll get back to this during the week! :smile:


molecule/resources/converge.yml
```diff
@@ -9,7 +9,9 @@
 
   pre_tasks:
     - name: Update apt cache
-      apt: update_cache=true cache_valid_time=600
+      apt:
+        update_cache: yes
+        cache_valid_time: 600
```
atb00ker, line 14: Not related, only for consistency.

tasks/apt.yml
```diff
@@ -1,5 +1,6 @@
 - name: Update APT package cache
-  apt: update_cache=yes
+  apt:
+    update_cache: yes
```
atb00ker, line 3: Not related, only for consistency.

tasks/freeradius.yml
```diff
@@ -0,0 +1,170 @@
...
+
+- name: delete_old_users
+  when: openwisp2_radius
+  cron:
+    name: delete_old_users
+    day: "*"
+    hour: "03"
+    minute: "30"
+    job: "/opt/openwisp2/env/bin/python /opt/openwisp2/manage.py delete_old_users"
+  tags: [openwisp2, radius]
```
atb00ker, line 170: Can be simply removed by using celery-beat.

templates/openwisp2/settings.py
```diff
@@ -91,15 +100,32 @@
...
+{% endif %}
+
 ROOT_URLCONF = 'openwisp2.urls'
 
 CHANNEL_LAYERS = {
     'default': {
-        "BACKEND": "channels_redis.core.RedisChannelLayer",
-        "CONFIG": {"hosts": [('localhost', 6379)]},
+        'BACKEND': 'channels_redis.core.RedisChannelLayer',
+        'CONFIG': {'hosts': [('localhost', 6379)]},
```
atb00ker, line 125: Not related, only for consistency.

templates/openwisp2/settings.py
```diff
@@ -132,17 +158,17 @@
 # FOR DJANGO REDIS
 
 CACHES = {
-    "default": {
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": "{{ openwisp2_redis_cache_url }}",
-        "OPTIONS": {
-            "CLIENT_CLASS": "django_redis.client.DefaultClient",
+    'default': {
```
atb00ker, line 161: Not related, only for consistency.

atb00ker (commented):
Started with openwisp_radius ansible setup. :smile:

tasks/freeradius.yml
nemesisdesign: cool, we should be able to add celery beat more easily once #206 and #215 are merged

defaults/main.yml
```diff
@@ -1,8 +1,10 @@
 openwisp2_python: python3
 ansible_python_interpreter: /usr/bin/python3
 openwisp2_network_topology: false
+openwisp2_radius: false
 openwisp2_controller_version: "0.7.post1"
 openwisp2_network_topology_version: "0.4"
+openwisp2_radius_version: "0.1"
```
nemesisdesign, line 7: I will put effort in speeding up the current work on openwisp-radius and release a 0.2 version so we can start with that one

defaults/main.yml
```diff
@@ -86,3 +89,24 @@ openwisp2_celery_worker_prefetch_multiplier: 1
...
+freeradius_sites_available_dir: "{{ freeradius_dir }}/sites-available"
+freeradius_sites_enabled_dir: "{{ freeradius_dir }}/sites-enabled"
+freeradius_sql:
+    driver: rlm_sql_postgresql
+    dialect: postgresql
+    host: localhost
+    port: 5432
+    dbname: freeradius
+    user: admin
+    password: admin
```
nemesisdesign, line 104: the default installation uses sqlite

defaults/main.yml
```diff
@@ -86,3 +89,24 @@ openwisp2_celery_worker_prefetch_multiplier: 1
...
+freeradius_rest:
+    url: "https://{{ inventory_hostname }}/api/v1/freeradius"
+openwisp2_users_auth_api: true
+openwisp2_radius_sms_backend: "sendsms.backends.console.SmsBackend"
+openwisp2_radius_extra_nas_types: >
+    [('cisco', 'Cisco Router')]
```
nemesisdesign, line 110: this won't be needed

templates/openwisp2/urls.py
```diff
@@ -16,6 +16,11 @@
     {% if openwisp2_network_topology %}
     url(r'^', include('openwisp_network_topology.urls')),
     {% endif %}
+    {% if openwisp2_radius %}
+        url(r'^', include('openwisp_radius.urls')),
+        url(r'^api/v1/', include('openwisp_utils.api.urls')),
+        url(r'^api/v1/', include('openwisp_users.api.urls')),
```
nemesisdesign, line 22: I think these two should be added independently from this if and should be always added, but I'm not sure if it's being done in other PRs

nemesisdesign (commented):
Thanks for starting this Ajay, great progress, I left some comments, more review and testing will be needed but it's surely going well.

defaults/main.yml
atb00ker: Yes, I have to work to work with SQLite, In fact I wanted to ask, is SQLite practical? Do people want that? Asking because sometimes I have noticed it has different behavior than PostgreSQL, while it makes sense for develop but just wondering if people need it in production as well!

templates/openwisp2/urls.py
atb00ker: Makes sense! :smile:

defaults/main.yml
nemesisdesign: Postgresql is better but the initial assumption is that ansible-openwisp2 would work as a base for people to get started and then they'd configure it as they preferred. Most people will use it with the default setting (sqlite), which will work for most cases.
Advanced cases will require some work on their side to set up either MySQL or PostgreSQL, depending on which one they prefer, I think this is ok for now, especially if we consider that in the near future we'll be migrating to docker, which if I'm not mistaken uses PG by default, right?

atb00ker: Yes, Got it.
docker-openwisp uses Postgresql by default.

README.md
```diff
@@ -401,88 +405,6 @@ username: admin
 password: admin
 ```
 
-Enabling the network topology module
```
atb00ker, line 404: @pandafy Was there a reason why the following was repeated? I have added just one line above, I think that should do the job, \imo repeating the same setup steps isn't very dry and causes confusion! :smile:

README.md
```diff
@@ -726,6 +654,33 @@ Below are listed all the variables you can customize (you may also want to take
...
+        dialect: sqlite
+        host: ""
+        port: ""
+        dbname: ""
+        user: ""
+        password: ""
+    freeradius_rest:
+        url: "https://{{ inventory_hostname }}/api/v1/freeradius"
+    freeradius_clients_ip: "0.0.0.0/0"
+    freeradius_clients_key: "admin"
```
atb00ker, line 677: @nemesisdesign are these defaults fine or can you suggest between defaults for `client.conf`?

atb00ker (commented):
~~@pandafy @nemesisdesign, please find the comments below, I think this out of the oven, a second pair eyes would be very helpful! :smile:~~ 
New error found, working on it.

templates/openwisp2/urls.py
```diff
@@ -19,6 +19,11 @@
     {% if openwisp2_firmware_upgrader %}
     url(r'^', include('openwisp_firmware_upgrader.urls')),
     {% endif %}
+    {% if openwisp2_radius %}
+    url(r'^', include('openwisp_radius.urls')),
+    url(r'^api/v1/', include('openwisp_users.api.urls')),
+    url(r'^api/v1/', include('openwisp_utils.api.urls')),
+    {% endif %}
```
atb00ker, line 26: I did not take out the api out of the if condition because `rest_framework.authtoken` needs to be added to openwisp-utils to be used with openwisp-radius, we can do it in https://github.com/openwisp/ansible-openwisp2/issues/234 separately, what do you think?

templates/openwisp2/settings.py
```diff
@@ -332,7 +387,7 @@
 # HTML minification with django pipeline
 PIPELINE = { 'PIPELINE_ENABLED': True }
 # static files minification and invalidation with django-compress-staticfiles
-STATICFILES_STORAGE = 'compress_staticfiles.storage.CompressStaticFilesStorage'
+STATICFILES_STORAGE = 'openwisp2.storage.OpenwispCompressStaticFilesStorage'
```
atb00ker, line 390: This was giving an error for openwisp-radius! 😄

atb00ker:
So, while, CI is failing again and again, it's working on my end, can someone confirm if this is working or not? :smile:

atb00ker:
@pandafy, stuck at https://github.com/openwisp/ansible-openwisp2/pull/223#issuecomment-739254882, if you have time, please look into it.
It works only device but running on molecule fails, looks like a django / dependency related thing.

templates/openwisp2/settings.py
pandafy: @atb00ker this is occurring becausue https://github.com/openwisp/openwisp-radius/commit/0819e35bef389b56b12fbba0ab85d9a679020106 was added after version 0.1.0. This should not occur if you test with the latest master. I guess we will require issuing 0.2.0 release of openwisp-radius.

pandafy:
I have not been able to pinpoint the exact reason for the failing molecule tests but I have narrowed down the cause. 

[cryptography 3.3 was released yesterday](https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#33---2020-12-08), which seems to not work with pyopenSSL==19.0.0. It works fine with pyopenSLL~=19.1.0. Changelog of `cryptography` does not mention anything about it and I could not find any other document to support this. 

I generated the dependency graph of openwisp-controller using `pipdeptree` that showed that pyOpenSSL==19.0.0 is getting installed. Another thing to mention is that `django-x509==0.9.1` pins pyOpenSSL to `<20.0.0`. I don't see any other module that pins the version of `pyopenSSL`


[Text  dependency graph of virtual environment of remote machine](https://pastebin.com/raw/XSiqCCQn)

nemesisdesign:
I wonder why we have a task that installs cryptography in this role, cryptography should be a dependency of django-x509 and I don't think we have to explicitly install it.

I am releasing a new minor version of django-x509 to ensure the cryptography dependency is bound: https://github.com/openwisp/django-x509/pull/108
Then I will remove the pip install task from this role.

atb00ker:
@pandafy actually, I was talking about :

```
  TASK [ansible-openwisp2 : migrate] *********************************************
  fatal: [openwisp2-ubuntu1804]: FAILED! => changed=false 
    cmd: ./manage.py migrate --noinput
    msg: |2-
    
      :stderr: Traceback (most recent call last):
        File "./manage.py", line 10, in <module>
          execute_from_command_line(sys.argv)
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
          utility.execute()
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/core/management/__init__.py", line 377, in execute
          django.setup()
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/__init__.py", line 24, in setup
          apps.populate(settings.INSTALLED_APPS)
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/apps/registry.py", line 122, in populate
          app_config.ready()
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/admin/apps.py", line 24, in ready
          self.module.autodiscover()
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/admin/__init__.py", line 24, in autodiscover
          autodiscover_modules('admin', register_to=site)
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/utils/module_loading.py", line 47, in autodiscover_modules
          import_module('%s.%s' % (app_config.name, module_to_search))
        File "/opt/openwisp2/env/lib/python3.6/importlib/__init__.py", line 126, in import_module
          return _bootstrap._gcd_import(name[level:], package, level)
        File "<frozen importlib._bootstrap>", line 994, in _gcd_import
        File "<frozen importlib._bootstrap>", line 971, in _find_and_load
        File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
        File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
        File "<frozen importlib._bootstrap_external>", line 678, in exec_module
        File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
        File "/opt/openwisp2/env/lib/python3.6/site-packages/openwisp_radius/admin.py", line 362, in <module>
          class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
        File "/opt/openwisp2/env/lib/python3.6/site-packages/openwisp_radius/admin.py", line 391, in RadiusBatchAdmin
          class Media:
        File "/opt/openwisp2/env/lib/python3.6/site-packages/openwisp_radius/admin.py", line 394, in Media
          static('openwisp-radius/js/strategy-switcher.js'),
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/templatetags/static.py", line 167, in static
          return StaticNode.handle_simple(path)
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/templatetags/static.py", line 118, in handle_simple
          return staticfiles_storage.url(path)
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/staticfiles/storage.py", line 147, in url
          return self._url(self.stored_name, name, force)
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/staticfiles/storage.py", line 126, in _url
          hashed_name = hashed_name_func(*args)
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/staticfiles/storage.py", line 418, in stored_name
          cache_name = self.clean_name(self.hashed_name(name))
        File "/opt/openwisp2/env/lib/python3.6/site-packages/django/contrib/staticfiles/storage.py", line 88, in hashed_name
          raise ValueError("The file '%s' could not be found with %r." % (filename, self))
      ValueError: The file 'openwisp-radius/js/strategy-switcher.js' could not be found with <openwisp2.storage.OpenwispCompressStaticFilesStorage object at 0x7f20a86b5c18>.
    path: /opt/openwisp2/env/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
    syspath:
    - /tmp/ansible_django_manage_payload_owop95xt/ansible_django_manage_payload.zip
    - /usr/lib/python36.zip
    - /usr/lib/python3.6
    - /usr/lib/python3.6/lib-dynload
    - /usr/local/lib/python3.6/dist-packages
```

pandafy:
@atb00ker this is occurring becausue openwisp/openwisp-radius@0819e35 was added after version 0.1.0. This should not occur if you test with the latest master. I guess we will require issuing 0.2.0 release of openwisp-radius

atb00ker:
@pandafy got it, since radius 2 release is so close, then let's wait for it! :smile:

nemesisdesign:
> @pandafy got it, since radius 2 release is so close, then let's wait for it!

@atb00ker openwisp-radius 0.2 is out! :grin: 
https://github.com/openwisp/openwisp-radius/releases/tag/0.2.0

atb00ker:
This one is ready! :smile:

nemesisdesign:
> This one is updated / ready!

Thanks! Give me some time to test it on some instances I am deploying, so I will be able to find out if there's any shortcoming that needs to be addressed

README.md
nemesisdesign: These sections are linked from other modules (topology and fw upgrader), they're here so we can quickly show to users interested in particular modules how to deploy them. I would not touch this section now and I would maintain consistency.

Modifying this section will need to update the links that point here and it would be better to deal with it in a different effort, making sure openwisp-radius works should be the priority now.

tasks/freeradius-mysql.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Freeradius additional mysql system packages
+  when: openwisp2_radius and freeradius_sql.dialect == "mysql"
```
nemesisdesign, line 3: the `when` statements in this file look redundant, because it's also stated in the `import_tasks`, which should be enough, isn't it?

tasks/freeradius-postgresql.yml
```diff
@@ -0,0 +1,43 @@
+---
+- name: Freeradius additional postgresql system packages
+  when: openwisp2_radius and freeradius_sql.dialect == "postgresql"
```
nemesisdesign, line 3: the `when` statements in this file look redundant, because it's also stated in the `import_tasks`, which should be enough, isn't it?

tasks/freeradius-mysql.yml
```diff
@@ -0,0 +1,38 @@
...
+  apt:
+    name:
+      - freeradius-mysql
+      - mariadb-server
```
nemesisdesign, line 7: installing MySQL server is out of scope for this role and should be done with another role which can handle its configuration.
It's not assured the DB will run on the same host.
Keep also n mind MariaDB is not MySQL although it was forked from it, so if mysql is specified, mysql should be used, mariadb has different DB drivers.

tasks/freeradius-mysql.yml
```diff
@@ -0,0 +1,38 @@
...
+    state: latest
+  notify: start mysql
+
+- name: Install pymysql
+  when: openwisp2_radius and freeradius_sql.dialect == "mysql"
+  pip:
+    name: pymysql
```
nemesisdesign, line 14: what is `pymysql` for? Is it for django or for freeradius? I don't think freeradius uses pymysql.

tasks/freeradius-mysql.yml
```diff
@@ -0,0 +1,38 @@
...
+  notify: start mysql
+
+- name: start mysql for migration
+  when: openwisp2_radius and freeradius_sql.dialect == "mysql"
+  meta: flush_handlers
+
+- name: Create freeradius database
+  when: openwisp2_radius and freeradius_sql.dialect == "mysql"
+  mysql_db:
+    login_unix_socket: /var/run/mysqld/mysqld.sock
```
nemesisdesign, line 29: doesn't this imply a limitation that the DB server must be on the same host?

README.md
```diff
@@ -579,12 +501,17 @@ Below are listed all the variables you can customize (you may also want to take
...
     openwisp2_network_topology: false
-    openwisp2_network_topology_version: "0.4"
+    openwisp2_network_topology_version: "0.5.1"
     openwisp2_firmware_upgrader: false
     openwisp2_firmware_upgrader_version: "0.1"
+    openwisp2_radius_version: "0.1"
+    # Enable the modules you want to use
+    openwisp2_network_topology: true
+    openwisp2_firmware_upgrader: true
+    openwisp2_radius: true
```
nemesisdesign, line 514: what about putting the default values here? Maybe that's what users expect and probably we should (after this PR) update this section to make it explicit.
I'm not 100% sure of it, what do you think? I fear this section may be getting confusing.

defaults/main.yml
```diff
@@ -92,3 +95,34 @@ openwisp2_celery_task_acks_late: true
 openwisp2_celery_broker_max_tries: 10
 openwisp2_django_celery_logging: false
 postfix_smtpd_relay_restrictions_override: "permit_sasl_authenticated, permit_mynetworks, check_relay_domains, reject_unauth_destination, reject"
+openwisp2_users_auth_api: true
+openwisp2_radius_sms_backend: "sendsms.backends.console.SmsBackend"
```
nemesisdesign, line 99: we'll need a documentation section in prose mentioning that this is the default value and if user needs to enable this feature they must do X, Y and Z, if you could start a draft regarding this I'll help to fill the blanks.

defaults/main.yml
```diff
@@ -92,3 +95,34 @@ openwisp2_celery_task_acks_late: true
...
+freeradius_mods_available_dir: "{{ freeradius_dir }}/mods-available"
+freeradius_mods_enabled_dir: "{{ freeradius_dir }}/mods-enabled"
+freeradius_sites_available_dir: "{{ freeradius_dir }}/sites-available"
+freeradius_sites_enabled_dir: "{{ freeradius_dir }}/sites-enabled"
+freeradius_sql:
+    driver: rlm_sql_sqlite
+    dialect: sqlite
+    host: ""
+    port: ""
+    dbname: ""
```
nemesisdesign, line 116: shouldn't this point to the openwisp DB? That's where the freeradius tables will be created by default.

handlers/main.yml
```diff
@@ -1,14 +1,38 @@
 ---
 
 - name: reload systemd
-  systemd: daemon_reload=yes
+  systemd:
+    daemon_reload: yes
```
nemesisdesign, line 5: the linter is complaining that it should be `true` rather than `yes`.

handlers/main.yml
```diff
@@ -1,14 +1,38 @@
...
+
+- name: start mysql
+  service:
+    name: mysql
+    state: started
+
+- name: start postgresql
+  service:
+    name: postgresql
+    state: started
```
nemesisdesign, line 24: why these handlers? Database do not need to be restarted since we do not touch their configurations, moreover, as I already stressed, the database management should be the concern of another role.

I understand you want to try to make it easier for the everyday user, but the everyday user who doesn't want to invest time in doing an advanced OpenWISP deployment should stick to sqlite, otherwise we'll end up having to give support to casual users in how to configure and manage their RDBMS which we surely don't have the resources to do.

handlers/main.yml
```diff
@@ -1,14 +1,38 @@
...
 
 - name: start redis
-  service: name=redis state=started
+  service:
+    name: redis
+    state: started
+
+- name: update-ca-certificates
+  shell: /usr/sbin/update-ca-certificates
```
nemesisdesign, line 32: try to use the command module, which is more limited than shell and hence less prone to security issues: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html
If `command` works, we should use it instead of shell.

tasks/apt.yml
```diff
@@ -1,7 +1,8 @@
 ---
 
 - name: Update APT package cache
-  apt: update_cache=yes
+  apt:
+    update_cache: yes
```
nemesisdesign, line 5: `true` to fix the warnings?

tasks/freeradius-mysql.yml
nemesisdesign: keep in mind that the python DB drivers are already installed in the `pip.yml` tasks:
https://github.com/openwisp/ansible-openwisp2/blob/694e7198bfa3dbb8ca42d168a46c3a84af4fe21d/tasks/pip.yml#L210-L236

tasks/freeradius-postgresql.yml
```diff
@@ -0,0 +1,43 @@
...
+  apt:
+    name:
+      - postgresql
```
nemesisdesign, line 6: same here for the topic of installing and managing the DB server.

tasks/freeradius-postgresql.yml
```diff
@@ -0,0 +1,43 @@
...
+      - freeradius-postgresql
+      - libpq-dev
+    state: latest
+  notify: start postgresql
+
+- name: Install psycopg2
+  when: openwisp2_radius and freeradius_sql.dialect == "postgresql"
+  pip:
+    name: psycopg2
```
nemesisdesign, line 15: Redundant with https://github.com/openwisp/ansible-openwisp2/blob/694e7198bfa3dbb8ca42d168a46c3a84af4fe21d/tasks/pip.yml#L210-L236

tasks/main.yml
```diff
@@ -23,6 +23,9 @@
 - import_tasks: django.yml
   tags: [openwisp2, django]
 
+- import_tasks: freeradius.yml
+  tags: [openwisp2, freeradius]
```
nemesisdesign, line 27: can you add a var to disable this? In case freeradius is installed on a separate server.

tasks/pip.yml
```diff
@@ -136,6 +143,20 @@
...
   until: result is success
 
+- name: Install openwisp2_radius and its dependencies
+  when: openwisp2_radius
+  pip:
+    name: "openwisp-radius~={{ openwisp2_radius_version }}"
+    state: latest
+    virtualenv: "{{ virtualenv_path }}"
+    virtualenv_python: "{{ openwisp2_python }}"
+    virtualenv_site_packages: yes
```
nemesisdesign, line 153: `true` to get rid of the warnings?

templates/openwisp2/urls.py
nemesisdesign: hopefully now that https://github.com/openwisp/ansible-openwisp2/commit/694e7198bfa3dbb8ca42d168a46c3a84af4fe21d is merged we shouldn't need this, right?

nemesisdesign (changes requested):
Great work @atb00ker! :+1:  I have started to review and test this.

Here's my first review (haven't tried to run it but will do soon).

Regarding the issue of managing the RDBMS, I am not very in favour of this for the reasons I explain in the comments below, however, I understand what you're trying to do, so we may as well try it, if we manage to do it we may help some users to deploy OpenWISP in a better way more easily, but we have to insert ways to disable this feature since until now I always recommended (and implemented) the installation and the management of the configurations of the RDBMS via separate roles, keep also in mind the DB may be deployed on a different VM or may be delegated to a hosting provider (eg: digitalocean) altogether, so it should be possible to disable all the tasks related to installing the DB server, creating the DB, etc.

The install of the DB drivers for python is ok, but is already performed in https://github.com/openwisp/ansible-openwisp2/blob/694e7198bfa3dbb8ca42d168a46c3a84af4fe21d/tasks/pip.yml#L210-L236.

If we go for the route of installing the DB though, the tasks which deal with the DB should not be freeradius specific but general, for the whole playbook, because the DB used by freeradius is exactly the same DB used by OpenWISP. Was this clear?

README.md
nemesisdesign: replied regarding clients elsewhere

templates/freeradius/sql_counter.j2
```diff
@@ -0,0 +1,41 @@
...
+
+    counter_name = Max-All-Session-Time
+    check_name = Max-All-Session
+    key = User-Name
+    reset = never
+
+    $INCLUDE ${modconfdir}/sql/counter/${dialect}/${.:instance}.conf
+}
+
+# The dailybandwidthcounter is added for django-freeradius
```
nemesisdesign, line 30: django-freeradius > openwisp-radius

templates/freeradius/sql_counter.j2
```diff
@@ -0,0 +1,41 @@
...
+   counter_name = Max-Daily-Session-Traffic
+   check_name = Max-Daily-Session-Traffic
+   sql_module_instance = sql
+   key = 'User-Name'
+   reset = daily
+   query = "SELECT SUM(acctinputoctets + acctoutputoctets) \
+            FROM radacct \
+            WHERE UserName='%{${key}}' \
+            AND UNIX_TIMESTAMP(acctstarttime) + acctsessiontime > '%%b'"
+}
```
nemesisdesign, line 41: These two are missing:

```
sqlcounter monthlycounter {
  sql_module_instance = sql
  dialect = postgresql

  counter_name = Monthly-Session-Time
  check_name = Max-Monthly-Session
  reply_name = Session-Timeout
  key = User-Name
  reset = monthly

  $INCLUDE ${modconfdir}/sql/counter/${dialect}/${.:instance}.conf
}

#  Set an account to expire T seconds after first login.
#  Requires the Expire-After attribute to be set, in seconds.
#  You may need to edit raddb/dictionary to add the Expire-After
#  attribute.
sqlcounter expire_on_login {
  sql_module_instance = sql
  dialect = postgresql

  counter_name = Expire-After-Initial-Login
  check_name = Expire-After
  key = User-Name
  reset = never

  $INCLUDE ${modconfdir}/sql/counter/${dialect}/${.:instance}.conf
}
```

templates/openwisp2/settings.py
```diff
@@ -110,12 +116,28 @@
...
+REST_AUTH_SERIALIZERS = {
+    'PASSWORD_RESET_SERIALIZER': 'openwisp_radius.api.serializers.PasswordResetSerializer',
+}
+REST_AUTH_REGISTER_SERIALIZERS = {
+    'REGISTER_SERIALIZER': 'openwisp_radius.api.serializers.RegisterSerializer',
+}
+OPENWISP_RADIUS_SMS_TOKEN_MAX_IP_DAILY = {{ openwisp2_radius_sms_token_max_ip_daily }}
+SENDSMS_BACKEND = '{{ openwisp2_radius_sms_backend }}'
+
+OPENWISP_USERS_AUTH_API = {{ openwisp2_users_auth_api }}
```
nemesisdesign, line 132: this one can be moved outside of this if since it can be used also by other modules.

nemesisdesign (changes requested):
More comments.

atb00ker:
Thanks for the review, I'll get back to them in some time! :smile:

atb00ker:
The changes so far look good, I'll have to give them more time though,

I agree about removing the database part which is being commented out right now, I'll remove it asap! :smile:

nemesisdesign (commented):
I rebased this branch on the current master.

templates/freeradius/sql.j2
```diff
@@ -0,0 +1,41 @@
...
+{% if freeradius_sql.dialect in ['postgresql', 'mysql'] %}
+	radius_db = "{{ openwisp2_database.name }}"
+  host = "{{ openwisp2_database.host }}"
+  port = "{{ openwisp2_database.port }}"
+  user = "{{ openwisp2_database.user }}"
+  password = "{{ openwisp2_database.password }}"
+{% elif 'sqlite' == freeradius_sql.dialect %}
+	sqlite {
+		filename = "{{ openwisp2_database.name }}"
+	}
```
atb00ker, line 13: I am getting:
```
rlm_sql_sqlite: libsqlite version: 3.27.2
   sqlite {
        filename = "/opt/openwisp2/db.sqlite3"
        busy_timeout = 200
   }
rlm_sql_sqlite: Database exists, but couldn't be opened: Permission denied
/etc/freeradius/3.0/mods-enabled/sql[1]: Instantiation failed for module "sql"
```

on this line.
I don't understand, it was working last time I checked. 
Was some setting for this changed recently? or something I am doing incorrectly? What about the version is `3.27.2` used by you as well? @nemesisdesign

atb00ker (commented):
Hi @nemesisdesign made some changes and rebased with master but it's not working for me anymore! I've mentioned the problem below I'll check it out later!

tasks/freeradius-mysql.yml
atb00ker: I add it because of a habit, because ansible changed it's behavior of `when` some versions ago, but I've removed it here because it seems to work! 😄

tasks/freeradius-postgresql.yml
atb00ker: I add it because of a habit, because ansible changed it's behavior of `when` some versions ago, but I've removed it here because it seems to work! 😄

tasks/freeradius-mysql.yml
atb00ker: Removed

tasks/freeradius-mysql.yml
atb00ker: Removed file! :-)

defaults/main.yml
atb00ker: It was already doing that, [it's failing right now](https://github.com/openwisp/ansible-openwisp2/pull/223#discussion_r550759166), need to investigate.

handlers/main.yml
atb00ker: Moved to command for now, but [unable to test it right now](https://github.com/openwisp/ansible-openwisp2/pull/223#discussion_r550759166). If it doesn't work after that problem is fixed, I'll move it back to `shell`.

tasks/freeradius.yml
```diff
@@ -0,0 +1,122 @@
...
+  when: openwisp2_radius and freeradius_sql.dialect == "mysql"
+
+- import_tasks: freeradius-postgresql.yml
+  when: openwisp2_radius and freeradius_sql.dialect == "postgresql"
+
+- name: Radius configurations
+  when: openwisp2_radius
+  template:
+    src: freeradius/radiusd.conf.j2
+    dest: "{{ freeradius_dir }}/radiusd.conf"
```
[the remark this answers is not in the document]
atb00ker: Removed! :-)

templates/freeradius/rest.j2
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    accounting {
+        uri = "${..connect_uri}/accounting/"
+        method = 'post'
+        body = 'json'
+        data = '{"status_type": "%{Acct-Status-Type}", "session_id": "%{Acct-Session-Id}", "unique_id": "%{Acct-Unique-Session-Id}", "username": "%{User-Name}", "realm": "%{Realm}", "nas_ip_address": "%{NAS-IP-Address}", "nas_port_id": "%{NAS-Port}", "nas_port_type": "%{NAS-Port-Type}", "session_time": "%{Acct-Session-Time}", "authentication": "%{Acct-Authentic}", "input_octets": "%{Acct-Input-Octets}", "output_octets": "%{Acct-Output-Octets}", "called_station_id": "%{Called-Station-Id}", "calling_station_id": "%{Calling-Station-Id}", "terminate_cause": "%{Acct-Terminate-Cause}", "service_type": "%{Service-Type}", "framed_protocol": "%{Framed-Protocol}", "framed_ip_address": "%{Framed-IP-Address}"}'
+        tls = ${..tls}
+    }
+}
```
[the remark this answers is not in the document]
atb00ker: Added documentation in [README](https://github.com/openwisp/ansible-openwisp2/pull/223/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R513)

templates/freeradius/sql_counter.j2
atb00ker: Added both of them! 😄

templates/freeradius/sql.j2
nemesisdesign: look for the task `set permissions to sqlite db`, the sqlite DB by default is readable only by the user www-data and the group www-data.

A possible solution may be to add the freeradius user to the www-data group.

tasks/freeradius.yml
```diff
@@ -0,0 +1,122 @@
...
+#   when: openwisp2_radius
+#   template:
+#     src: freeradius/clients.conf.j2
+#     dest: "{{ freeradius_dir }}/site"
+#     mode: 0640
+#     owner: freerad
+#     group: freerad
+#   notify: restart freeradius
+
+# - name: Remove unnecessary modules
```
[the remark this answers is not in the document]
nemesisdesign: ok, let's keep this in mind if we stumble upon the same issue again, EAP is worth fighting for.

nemesisdesign (commented):
Thanks @atb00ker, I updated the branch again (this time I used the merge) and will be testing it.

templates/freeradius/sql.j2
atb00ker: I think I made the db have 777 permissions when I was last testing, but I'll check again.
From what I remember the problem was because only one application can write sqlite db at a time and freeradius and the other one were fighting for the write permission.

However, I'll double check this.

nemesisdesign: I think is simply a permission issue, the sqlite DB is readable only by group and owner, the freeradius user is not owner nor group.

Yes, only one process can write to the DB at time, if more processes try to write while another process is writing they'll get an error, but that won't stop the other processes from reading the DB (the celery worker for example reads the DB while the django application also reads it, so freeradius can read it too).

The write operations are performed via the openwisp-radius API, hence only django will write to the DB, freeradius should not be writing to the DB directly unless I'm forgetting anything.

atb00ker: It's working! :smile:

templates/freeradius/openwisp_site.j2
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        Auth-Type MS-CHAP {
+                mschap
+        }
+
+        Auth-Type EAP {
+                eap
+        }
+    }
```
atb00ker, line 44: After reading: https://community.privacyidea.org/t/fresh-install-pid-but-i-get-auth-type-eap-section/1067/5

I searched for activating EAP Authentication module works, I'll look it but up but please check it out, my knowledge of freeradius is very limited and even small changes seem to break it completely.

atb00ker, line 44: After reading: https://community.privacyidea.org/t/fresh-install-pid-but-i-get-auth-type-eap-section/1067/5

I searched for activating EAP Authentication module works, I'll look it but up but please check it out, my knowledge of freeradius is very limited and even small changes seem to break it completely.

atb00ker (commented):
What remains for this PR:
1. EAP problem
2. https://github.com/openwisp/ansible-openwisp2/pull/223#discussion_r550754617
3. Testing

templates/openwisp2/settings.py
```diff
@@ -352,8 +352,7 @@
             'formatter': 'verbose',
             'filename': os.path.join(BASE_DIR, 'log/openwisp2.log'),
             'maxBytes': 15728640,
-            'backupCount': 3,
-            'formatter': 'verbose'
```
nemesisdesign, line 302: why this change here?

templates/freeradius/openwisp_site.j2
```diff
@@ -25,7 +25,23 @@ server default {
...
+        Auth-Type CHAP {
+                chap
+        }
+
+        Auth-Type MS-CHAP {
+                mschap
+        }
+
+        Auth-Type EAP {
+                eap
```
nemesisdesign, line 42: indentation looks inconsistent

templates/freeradius/openwisp_site.j2
nemesisdesign: EAP needs the inner-tunnel freeradius site as well if I'm not mistaken.
It will take me a while to find the time to test it, but I will do it, I must have the config somewhere.

nemesisdesign (commented):
> Unfortunately, I've not used SENDSMS_BACKEND option outside testing on console, so can you please share which backend is used most often with openwisp?
> I'll document accordingly! smile

Every installation I deployed this on uses a different one, some services have their own API or already have some kind of contract or agreement with a specific provider, usually I write dedicated sms backends in every case.

This means that the console backend is a good default.

README.md
```diff
@@ -579,12 +618,22 @@ Below are listed all the variables you can customize (you may also want to take
     - openwisp.openwisp2
   vars:
     # openwisp-controler version
-    openwisp2_controller_version: "0.4"
+    openwisp2_controller_version: "0.8.1"
```
nemesisdesign, line 621: the current default value is `0.8.2`

README.md
```diff
@@ -579,12 +618,22 @@ Below are listed all the variables you can customize (you may also want to take
...
     # optional openwisp2 modules
     openwisp2_network_topology: false
-    openwisp2_network_topology_version: "0.4"
+    openwisp2_network_topology_version: "0.5.1"
     openwisp2_firmware_upgrader: false
     openwisp2_firmware_upgrader_version: "0.1"
+    openwisp2_radius_version: "0.1"
```
nemesisdesign, line 627: this should be 0.2.1

handlers/main.yml
```diff
@@ -1,14 +1,28 @@
...
 
 - name: start redis
-  service: name=redis state=started
+  service:
+    name: redis
+    state: started
+
+- name: update-ca-certificates
+  command: /usr/sbin/update-ca-certificates
+  when: ansible_os_family == "Debian"
```
nemesisdesign, line 23: we shouldn't need this since we only support debian family OS now

molecule/resources/converge.yml
```diff
@@ -7,10 +7,13 @@
...
     openwisp2_firmware_upgrader: true
+    openwisp2_radius: true
 
   pre_tasks:
     - name: Update apt cache
-      apt: update_cache=true cache_valid_time=600
+      apt:
+        update_cache: true
+        cache_valid_time: 600
       when: ansible_os_family == 'Debian'
```
nemesisdesign, line 17: can you remove this as well please?

tasks/nginx.yml
```diff
@@ -26,6 +26,17 @@
...
 
+- name: Copy SSL cert to be added to trusted Cert (for freeradius)
+  copy:
+    src: "{{ openwisp2_ssl_cert }}"
+    dest: /usr/local/share/ca-certificates/openwisp-ssl-server.crt
+    remote_src: true
+    owner: "root"
+    group: "root"
+    mode: "0644"
+  when: ansible_os_family == 'Debian'
```
nemesisdesign, line 37: we shouldn't need this since we only support debian family OS now

templates/freeradius/sql.j2
```diff
@@ -0,0 +1,40 @@
+sql {
+  driver = "{{ freeradius_sql.driver }}"
+  dialect = "{{ freeradius_sql.dialect }}"
+{% if freeradius_sql.dialect in ['postgresql', 'mysql'] %}
+	radius_db = "{{ openwisp2_database.name }}"
```
nemesisdesign, line 5: are you indenting this line differently in order to get the indentation right in the resulting file?
If so, we shouldn't care about that, maintainability of this role for us is more important

templates/openwisp2/urls.py
```diff
@@ -17,15 +17,18 @@
     {% endif %}
```
nemesisdesign, line 17: since you're reindenting these lines, check the `if openwisp2_controller_urls` which was added recently

nemesisdesign (changes requested):
@atb00ker found some minor issues

templates/openwisp2/settings.py
atb00ker: `'formatter': 'verbose'` is repeated, it's not causing error so I thought it sneaked in with one of the recent PRs, I can't think of a reason why repeating this key would be useful.

nemesisdesign (commented):
The indentation in the freeradius config files is inconsistent, I'm fixing that, along some minor improvements.

I found one issue with the `dailybandwidthcounter`, see below.

Another thing: is there already a test in the build which checks if freeradius is correctly started?

README.md
```diff
@@ -483,6 +483,45 @@ you will update the above playbook as follows:
...
+please disable this option by setting it to `false`.
+
+**Step 5**: [Run the playbook](#run-the-playbook)
+
+When the playbook is done running, if you got no errors you can login at:
+
+    https://openwisp2.mydomain.com/admin
+    username: admin
+    password: admin
+
```
nemesisdesign, line 524: Slightly bad news, on Ubuntu 20 the freeradius  version installed by default has a bug in the rest module which doesn't make it work with HTTP2.

I just double checked the HTTP2 setting in this role and is disabled by default, but it may be good to take a note about this issue here. This problem should be fixed in freeradius 3.0.22 is released.

nemesisdesign (commented):
Found another issue, this one affecting PostgreSQL: https://github.com/FreeRADIUS/freeradius-server/issues/3867. I will send a workaround to fix this.

README.md
atb00ker: I've been using 3.0.20 on `docker-openwisp` for this reason.

atb00ker:
> Another thing: is there already a test in the build which checks if freeradius is correctly started?

No.

atb00ker:
> Another thing: is there already a test in the build which checks if freeradius is correctly started?

Here is an update, I came up with the following but it doesn't work because the IP address is not allowed to use freeradius, so I thought of https://github.com/openwisp/openwisp-radius/issues/229 , we will set `0.0.0.0/0` for testing and we should be good to go, how does this sound?

```
    - name: Check Freeradius
      block:
        - name: Get radius-token for admin user
          uri:
            url: "https://{{ inventory_hostname }}/api/v1/default/account/token/"
            validate_certs: false
            method: POST
            body_format: form-urlencoded
            body:
              username: admin
              password: admin

        - name: Check if freeradius is running
          command: radtest admin admin localhost 0 testing123
          register: freeradius_status
          failed_when: '"Received Access-Accept" not in freeradius_status.stdout'
```

nemesisdesign (commented):
> > Another thing: is there already a test in the build which checks if freeradius is correctly started?
> 
> Here is an update, I came up with the following but it doesn't work because the IP address is not allowed to use freeradius, so I thought of [openwisp/openwisp-radius#229](https://github.com/openwisp/openwisp-radius/issues/229) , we will set `0.0.0.0/0` for testing and we should be good to go, how does this sound?
> 
> ```
>     - name: Check Freeradius
>       block:
>         - name: Get radius-token for admin user
>           uri:
>             url: "https://{{ inventory_hostname }}/api/v1/default/account/token/"
>             validate_certs: false
>             method: POST
>             body_format: form-urlencoded
>             body:
>               username: admin
>               password: admin
> ```

Is this for testing or is it a prerequisite for the next test?
I think the former, do you confrim?
 
> ```
>         - name: Check if freeradius is running
>           command: radtest admin admin localhost 0 testing123
>           register: freeradius_status
>           failed_when: '"Received Access-Accept" not in freeradius_status.stdout'
> ```

They look good, I'm curious to see if it works.

atb00ker:
> I think the former, do you confrim?

@nemesisdesign  Yes, it's setup step, if the user does not have a radius-token, freeradius API cannot communicate. (this behavior seems a little wrong to me though..)

nemesisdesign:
> > I think the former, do you confrim?
> 
> @nemesisdesign Yes, it's setup step, if the user does not have a radius-token, freeradius API cannot communicate. (this behavior seems a little wrong to me though..)

oh right, it's because without it we are not able to associate a user to an org, unless we configure the freeradius site to have the organization uuid and token in its configuration, which would require to have 1 freeradius site for each org, which I pushed to avoid so that it would be easier to create new organizations (otherwise most users will just 1 org and there will be no point in implementing multi-tenancy).

atb00ker:
Update: Created PR https://github.com/openwisp/openwisp-radius/pull/232
Tested it on docker-openwisp here: https://github.com/openwisp/docker-openwisp/pull/143

nemesisdesign (commented):
@atb00ker https://github.com/openwisp/openwisp-radius/pull/232 is merged, thank you.

templates/freeradius/sqlcounter.j2
```diff
@@ -0,0 +1,65 @@
+# The dailycounter is included by default in the freeradius conf
+
+sqlcounter dailycounter {
+    sql_module_instance = sql
+    dialect = ${modules.sql.dialect}
```
atb00ker, line 5: New problem, tests failing because ` ${modules.sql.dialect}` is not found.
I tried `dialect = "{{ freeradius_sql.dialect }}"` and that worked. Not sure why the value isn't found in `sqlcounter` now but should I make this change?

templates/freeradius/sqlcounter.j2
```diff
@@ -0,0 +1,65 @@
...
+    key = User-Name
+    reset = daily
+
+    $INCLUDE ${modconfdir}/sql/counter/${dialect}/${.:instance}.conf
+}
+
+# The dailybandwidthcounter is added by OpenWISP
+sqlcounter dailybandwidthcounter {
+    sql_module_instance = sql
+    dialect = ${modules.sql.dialect}
```
atb00ker, line 20: New problem, tests failing because ` ${modules.sql.dialect}` is not found.
I tried `dialect = "{{ freeradius_sql.dialect }}"` and that worked. Not sure why the value isn't found in `sqlcounter` now but should I make this change?

templates/freeradius/sqlcounter.j2
```diff
@@ -0,0 +1,65 @@
...
+    check_name = Max-Daily-Session-Traffic
+    key = User-Name
+    reset = daily
+
+    $INCLUDE ${modconfdir}/sql/counter/${dialect}/${.:instance}.conf
+}
+
+sqlcounter noresetcounter {
+    sql_module_instance = sql
+    dialect = ${modules.sql.dialect}
```
atb00ker, line 32: New problem, tests failing because ` ${modules.sql.dialect}` is not found.
I tried `dialect = "{{ freeradius_sql.dialect }}"` and that worked. Not sure why the value isn't found in `sqlcounter` now but should I make this change?

templates/freeradius/sqlcounter.j2
```diff
@@ -0,0 +1,65 @@
...
+    check_name = Max-All-Session
+    key = User-Name
+    reset = never
+
+    $INCLUDE ${modconfdir}/sql/counter/${dialect}/${.:instance}.conf
+}
+
+sqlcounter monthlycounter {
+    sql_module_instance = sql
+    dialect = ${modules.sql.dialect}
```
atb00ker, line 44: New problem, tests failing because ` ${modules.sql.dialect}` is not found.
I tried `dialect = "{{ freeradius_sql.dialect }}"` and that worked. Not sure why the value isn't found in `sqlcounter` now but should I make this change?

templates/freeradius/sqlcounter.j2
```diff
@@ -0,0 +1,65 @@
...
+    reply_name = Session-Timeout
+    key = User-Name
+    reset = monthly
+
+    $INCLUDE ${modconfdir}/sql/counter/${dialect}/${.:instance}.conf
+}
+
+sqlcounter expire_on_login {
+    sql_module_instance = sql
+    dialect = ${modules.sql.dialect}
```
atb00ker, line 57: New problem, tests failing because ` ${modules.sql.dialect}` is not found.
I tried `dialect = "{{ freeradius_sql.dialect }}"` and that worked. Not sure why the value isn't found in `sqlcounter` now but should I make this change?

atb00ker (commented):
Found a new problem! :smile:

tasks/freeradius.yml
```diff
@@ -0,0 +1,156 @@
...
+    owner: freerad
+    group: freerad
+  notify: restart freeradius
+
+- name: Enable SQL module
+  file:
+    src: "{{ freeradius_mods_available_dir }}/sql"
+    # the "1_" prefix is put on purpose to force loading the sql module
+    # before the sqlcounter module to avoid an annoying startup error
+    dest: "{{ freeradius_mods_enabled_dir }}/1_sql"
```
nemesisdesign, line 38: @atb00ker ensure you have this file in your filesystem, the sql module must loaded before sqlcounter otherwise that error you mentioned appears. I thought I solved it with this change (I tested it).

templates/freeradius/sqlcounter.j2
nemesisdesign: Please see https://github.com/openwisp/ansible-openwisp2/pull/223/files#r573134118

tasks/freeradius.yml
atb00ker: Hmm, this solution didn't work for me and CI is failing too, I'll check it out, it should have worked though, I did something similar (renamed `sqlcounter` to `sql-counter`) in an earlier commit which worked.

nemesisdesign: ah ok, I did not understand that, please always leave comments in the code when you do these kind of things, if it worked we can restore it

templates/openwisp2/settings.py
```diff
@@ -162,9 +183,45 @@
...
     'delete_old_notifications': {
         'task': 'openwisp_notifications.tasks.delete_old_notifications',
-        'schedule': timedelta(days=1),
+        'schedule': crontab(**{ {{ cron_delete_old_notifications }} }),
         'args': ({{ openwisp2_notifications_delete_old_notifications }},),
     },
+{% if openwisp2_radius %}
+
+    'deactivate_expired_users': {
+        'task': 'openwisp_radius.tasks.cleanup_stale_radacct',
```
nemesisdesign, line 192: this looks wrong, since `openwisp_radius.tasks.cleanup_stale_radacct` is repeated below

nemesisdesign (changes requested):

.github/workflows/ci.yml
```diff
@@ -28,14 +28,9 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
 
-      - name: Set up Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: '3.x'
-
```
atb00ker, line 35: New problem with CI, had to remove this to fix it!

atb00ker: Argh! Will need update python version because of other errors in 3.6! :-/

atb00ker:
CI is failing because freeradius setup has some error, I'll continue to work on it tomorrow! :smile:

nemesisdesign (commented):
Hey @atb00ker, I forgot what's the status of this.. I know that on my side I want to make sure the inner tunnel configuration is correct so it can be used to provide WPA2 Enterprise, what about the rest? I see only Ubuntu 18 is failing, any idea why?

Anything else missing?

I have been using this branch to deploy a couple of instances with the radius module and they're working fine.

atb00ker:
> what about the rest? 

When openwisp-radius 3.0 is released, we need to enable test for it, that's it. :smile:

>  I see only Ubuntu 18 is failing, any idea why?

I have failed to reproduce it so far.
Do we want to put effort into it? 18.04 is now at EOL and only getting security updates!

> I have been using this branch to deploy a couple of instances with the radius module and they're working fine.

That's great news! :smile:

nemesisdesign (commented):
I added a basic inner tunnel configuration as well.
Let's encourage more people to test it.
However, I believe it's good enough for its first cycle of iterations.
PS: not sure why the last build failed, it's working in my tests, but will test it again locally asap to find out.

.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       - name: Install python dependencies
         run: |
-          sudo python3 -m pip install molecule[docker] yamllint ansible ansible-lint docker openwisp-utils[qa]
```
atb00ker: If I am not wrong, this will be required in the future though; direct use of pip will go away!

nemesisdesign: ok we can revert this change!

molecule/resources/verify.yml
```diff
@@ -64,3 +64,22 @@
         - name: Show OpenWisp log
           debug:
             var: openwisp_log
+
+    - name: Check Freeradius
+    # TODO: This test should work when openwisp-radius 0.3.0 is released!
```
atb00ker, line 69: This remains a blocker to merge this PR.
We would need to release the next version of openwisp-radius to enable testing here.

tasks/freeradius.yml
```diff
@@ -152,3 +151,12 @@
     owner: freerad
     group: freerad
   notify: restart freeradius
+
+- name: Inner tunnel
+  template:
+    src: freeradius/openwisp_site.j2
```
nemesisdesign: @atb00ker isn't this wrong?

molecule/resources/verify.yml
nemesisdesign: yes we're getting near to that :+1:

tasks/freeradius.yml
atb00ker: Yes, this was one of the causes of build failure,fixed it in the latest commit! :smile:

nemesisdesign (commented):
This is basically ready!
OpenWISP RADIUS 0.3 milestone issues: https://github.com/openwisp/openwisp-radius/milestone/1

nemesisdesign (commented):
I found an issue that prevents the group configurations (radius group check, radius group reply) from working when using registration with phone number because the plus sign gets escaped. This is breaking group authorization rules.

To fix it we need to add a configurable variable called `openwisp2_freeradius_safe_characters` defaulting to `+@abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.-_: /`.

Then in the configuration of the SQL dialect in use has to be changed, I think we have to replace any line which contains `safe_characters =`, because it seems the default is different depending on the DB:

- https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/raddb/mods-config/sql/main/postgresql/queries.conf here the default is commented out
- https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/raddb/mods-config/sql/main/mysql/queries.conf here it's not commented out

The value we put should be:

```
safe_characters = "{{ openwisp2_freeradius_safe_characters }}"
```

defaults/main.yml
```diff
@@ -104,8 +108,45 @@ openwisp2_celery_broker_url: redis://{{ openwisp2_redis_host }}:{{ openwisp2_red
...
+    dialect: "{{ freeradius_db_map[openwisp2_database.engine].dialect }}"
+freeradius_rest:
+    url: "https://{{ inventory_hostname }}/api/v1/freeradius"
+freeradius_expire_attr_after_seconds: 86400
+cron_delete_old_notifications: "'hour': 0, 'minute': 0"
+cron_deactivate_expired_users: "'hour': 0, 'minute': 5"
+cron_delete_old_users: "'hour': 0, 'minute': 10"
+cron_cleanup_stale_radacct: "'hour': 0, 'minute': 20"
+cron_delete_old_postauth: "'hour': 0, 'minute': 30"
+cron_delete_old_radacct: "'hour': 1, 'minute': 30"
```
nemesisdesign, line 152: I think we should prefix these cron variables with `openwisp2` for consistency, since it's configuration related to openwisp.

nemesisdesign:
> I found an issue that prevents the group configurations (radius group check, radius group reply) from working when using registration with phone number because the plus sign gets escaped. This is breaking group authorization rules.

Commit https://github.com/openwisp/ansible-openwisp2/pull/223/commits/8bebc284ef7d44ab7e10277c4e34a4cbcd3cb654 should fix it. I tested it. WIll keep an eye on the CI.

nemesisdesign:
I have merged this manually in the `dev` branch, I will keep this open until `dev` is ready to be recommended as a way to install OpenWISP RADIUS, something I'm working on with @pandafy.

atb00ker:
@nemesisdesign awesome, big win! 😄

PR #229, opened 2020-12-04
[fix] Install geographic libraries separately

nemesisdesign:
If the playbook is run the first time with a different DB configured
(eg: Postgis or MySQL), the geographic libraries won't be installed.
Therefore we need to install the geodjango dependencies in a step that
is separate than the installation of spatialite.

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -58,14 +58,22 @@
     line: 'bind 127.0.0.1'
     backrefs: yes
 
+- name: Install geographic libraries
+  apt:
+    name:
+      - gdal-bin
+      - libproj-dev
+      - libgeos-dev
+  retries: 5
+  delay: 10
+  register: result
+  until: result is success
+
 - name: Install spatialite
   when: openwisp2_database.engine == "django.contrib.gis.db.backends.spatialite"
   apt:
     name:
       - sqlite3
-      - gdal-bin
-      - libproj-dev
-      - libgeos-dev
       - libspatialite-dev
   retries: 5
   delay: 10


pandafy (commented):
The changes look good to me. 

According to the [Spatial Database section of GeoDjango installation documentation](https://docs.djangoproject.com/en/3.1/ref/contrib/gis/install/#spatial-database), `libproj-dev` is needed only for Postgres and SQLite. Will it be a good idea to install it only alongside SQLite by default?

nemesisdesign:
> The changes look good to me.
> 
> According to the [Spatial Database section of GeoDjango installation documentation](https://docs.djangoproject.com/en/3.1/ref/contrib/gis/install/#spatial-database), `libproj-dev` is needed only for Postgres and SQLite. Will it be a good idea to install it only alongside SQLite by default?

We'd need to add another if that checks if DB is postgis or spatialite.
But since size is about 1MB: https://packages.debian.org/stretch/libproj-dev, I'd just install it anyway to keep things simple.

atb00ker:
How are you guys reviewing the work here?
GitHub Actions has added a comment in every second line, I can't see the code properly, and I don't see a way to collapse these boxes. Am I missing something? :smile:

atb00ker (approved):

pandafy (approved):
> We'd need to add another if that checks if DB is postgis or spatialite.
But since size is about 1MB: https://packages.debian.org/stretch/libproj-dev, I'd just install it anyway to keep things simple.

LGTM! :+1:

nemesisdesign:
> ~How are you guys reviewing the work here?~
> ~GitHub Actions has added a comment in every second line, I can't see the code properly, and I don't see a way to collapse these boxes. Am I missing something? smile~
> 
> Never mind, those are just "Unchanged files with check annotations", I didn't realize that before!

I don't understand why that happens, but I'm ignoring those for now, although would be good to get rid of those warnings once the release is done.

atb00ker:
> I don't understand why that happens, but I'm ignoring those for now, although would be good to get rid of those warnings once the release is done.

I'll look into it! :smile:

PR #235, opened 2020-12-05
[fix] Main template on login page

### LINKED ISSUE #233: [bug] Main template not loding in /accounts/login/
PS: this happens on a fresh installation of ansible-openwisp2 with the latest patches merged. At first I thought I was able to replicate it in the dev env of openwisp-users but subsequent attempts do not show the issue, although confirmation would be welcome.

Try to access /accounts/login/.

Something like the following screenshot is shown:

Screenshot from 2020-12-03 17-52-31

But we expect to see the main theme as in the URL which is used to confirm the email address.

Screenshot from 2020-12-03 17-53-03

This bug is triggered after confirming an email address.

Purhan:
Fixes #233

### DIFF
--- a/templates/openwisp2/settings.py
+++ b/templates/openwisp2/settings.py
@@ -35,6 +35,9 @@ INSTALLED_APPS = [
     # overrides allauth templates
     # must precede allauth
     'openwisp_users.accounts',
+    # openwisp2 admin theme
+    # (must be loaded here)
+    'openwisp_utils.admin_theme',
     'allauth',
     'allauth.account',
     'allauth.socialaccount',
@@ -54,9 +57,6 @@ INSTALLED_APPS = [
     'openwisp_firmware_upgrader',
     'private_storage',
 {% endif %}
-    # openwisp2 admin theme
-    # (must be loaded here)
-    'openwisp_utils.admin_theme',
     # admin
     'django.contrib.admin',
     'django.forms',


Purhan:
Loading the admin template before Allauth seems to have fixed the issue

image

nemesisdesign (commented):
I see, but in this case, I'm not sure the notification widget works, can you check that please?

Purhan:
> I see, but in this case, I'm not sure the notification widget works, can you check that please?

You're right, this is a WIP in that case.

nemesisdesign:
@pandafy we may need to think about a solution for this problem: if we move the theme we fix the issue but break the notifications widget

pandafy:
After lots of experimenting with admin templates I found out that changing [`base_site.html` in openwisp_utils.admin_theme](https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/admin_theme/templates/admin/base_site.html) to `base.html` properly renders the `/accounts/login` page and the notification widget in the admin dashboard. 

I am not completely sure whether this is the right way to go or whether doing this breaks other things. We certainly have to test different modules to ensure that bringing out this change does not break their functionality. 

@Purhan will you be able to help me on this? [Overriding admin templates](https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#overriding-admin-templates) section in the Django documentation is a good place to start if this concept is new to you.

Purhan:
> I found out that changing base_site.html in openwisp_utils.admin_theme to base.html properly renders the /accounts/login page and the notification widget in the admin dashboard.

@pandafy Testing this on [openwisp-utils](https://github.com/openwisp/openwisp-utils/) master, it currently breaks the Menu icon on the admin dashboard.

image
image

Purhan:
Meanwhile, I found that installing the development version of [openwisp-users](https://github.com/openwisp/openwisp-users/), which now has support for `django-allauth==0.43.0`, completely fixes this issue. We could make a point release if that's a feasible option.

Purhan:
I'm afraid I'll have to bother you with this just one last time :sweat_smile:, 
> Testing this on openwisp-utils master, it currently breaks the Menu icon on the admin dashboard.

can you confirm this? Should I still open an issue on `openwisp-users`, and if so, any suggested description I should use for it?

nemesisdesign:
> Meanwhile, I found that installing the development version of [openwisp-users](https://github.com/openwisp/openwisp-users/), which now has support for `django-allauth==0.43.0`, completely fixes this issue. We could make a point release if that's a feasible option.

I confirm this, thank you @Purhan for pointing it out! I just tested it on an instance deployed with the latest ansible-openwisp2 and indeed it fixes the issue, so [I released openwisp-users 0.5.1](https://pypi.org/project/openwisp-users/0.5.1/) and I think no further action is needed, let's close this.

PR #238, opened 2020-12-08
[fix] Fixed leaflet marker icon #232

### LINKED ISSUE #232: [bug] Static minification breaks leaflet marker icon
On the current latest master.

How to replicate:

- create a device
- assign a map location to the device

Expected outcome: marker is shown
Actual outcome: marker icon is not found because its name is miscalculated

Screenshot from 2020-12-04 13-39-11

Screenshot from 2020-12-04 13-42-08

pandafy:
Bug:
HTTP requests to fetch the marker icon were bogus due to a bug in
leaflet that made it incapable to process versioned static files
created by ManifestFilesMixin.

Fix:
- A custom static storage class with option to exclude files based on
  Unix shell-style wildcards.
- PNG files are excluded from versioning

Fixes #232

### DIFF
--- a/files/storage.py
+++ b/files/storage.py
@@ -1,0 +1,33 @@
+import fnmatch
+
+from compress_staticfiles.storage import (
+    CompressStaticFilesStorage as BaseCompressStaticFilesStorage,
+)
+from django.conf import settings
+
+
+class FileHashedNameMixin:
+    default_excluded_patterns = ['*.png']
+    excluded_patterns = default_excluded_patterns + getattr(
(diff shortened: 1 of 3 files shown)


nemesisdesign (changes requested):
Great progress @pandafy!

Can we exclude by default only the leaflet related png?
Is good to let the storage handle all the other PNGs since they may be changed from time to time and we don't want to require users to clear their cache.

I also do not see the changed setting in settings.py which sets this storage, is it present in this PR? I don't see it

nemesisdesign (approved):
Tested and it works, great work @pandafy :+1: :+1:

PR #241, opened 2020-12-12
[feature] Add API URLs by default #234

### LINKED ISSUE #234: [feature] Add API URLs by default
In order to facilitate enabling the API docs and the openwisp-users API (which is a pre-requisite for other modules), we should do the following:

- Add `rest_framework.authtoken` and `drf_yasg` to `settings.INSTALLED_APPS`
- Add the following URLs to urls.py:
  - `url(r'^api/v1/', include('openwisp_utils.api.urls'))`
  - `url(r'^api/v1/', include('openwisp_users.api.urls'))`

The reasons for including these by default are the following:

- makes it easy to enable the API features by just changing a setting (if users want to do it with the current master they have to do the steps above on their own, which will cause them most probably to not be able to do it and will write to us to know how to do it)
- the openwisp-radius PR #223 is adding these with a conditional depending if openwisp-radius is enabled, but these endpoints are used also for other modules
- in the near future we may switch to enable the API features by default

I see no downside in doing this since even if the URLs are loaded they won't be available if the API settings are not set to `True`.

@atb00ker @pandafy @NoumbissiValere what do you think?

Purhan:
Closes #234

### DIFF
--- a/templates/openwisp2/settings.py
+++ b/templates/openwisp2/settings.py
@@ -57,6 +57,8 @@ INSTALLED_APPS = [
     # openwisp2 admin theme
     # (must be loaded here)
     'openwisp_utils.admin_theme',
+    'rest_framework.authtoken',
+    'drf_yasg',
     # admin
     'django.contrib.admin',
     'django.forms',
--- a/templates/openwisp2/settings.py
+++ b/templates/openwisp2/settings.py
@@ -57,8 +57,6 @@ INSTALLED_APPS = [
     # openwisp2 admin theme
     # (must be loaded here)
     'openwisp_utils.admin_theme',
-    'rest_framework.authtoken',
-    'drf_yasg',
     # admin
     'django.contrib.admin',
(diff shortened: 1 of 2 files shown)


templates/openwisp2/settings.py
```diff
@@ -57,6 +57,8 @@
     # openwisp2 admin theme
     # (must be loaded here)
     'openwisp_utils.admin_theme',
+    'rest_framework.authtoken',
+    'drf_yasg',
```
nemesisdesign, line 61: I would move these two below, after `rest_framework_gis`

nemesisdesign (changes requested):
Looks good, will check what's wrong with the build later, in the meanwhile see my comment below please.

templates/openwisp2/settings.py
Purhan: Done

nemesisdesign (approved):
Thanks!

PR #244, opened 2020-12-22
[change] Edit permissions of /opt/openwisp2 #243

### LINKED ISSUE #243: [change] Edit permissions of /opt/openwisp2 and settings.py
Make `/opt/openwisp2` and any directory created under it not readable, nor executable except for its owner and its group.
Ensure `/opt/openwisp2/openwisp2/settings.py` is only readable by owner and group.

Purhan:
Closes #243

### DIFF
--- a/tasks/django.yml
+++ b/tasks/django.yml
@@ -156,3 +156,12 @@
     or "changed" in load_initial_data_result.stdout
   args:
     chdir: "{{ openwisp2_path }}"
+
+- name: set permissions to "{{ openwisp2_path }}"
+  file:
+    path: "{{ openwisp2_path }}"
+    state: directory
+    group: "{{ www_group }}"
+    owner: "{{ www_user }}"
+    mode: 0770
+    recurse: yes


Purhan:
Permissions for `/opt/openwisp2/openwisp2/settings.py`
image

tasks/django.yml
```diff
@@ -156,3 +156,12 @@
     or "changed" in load_initial_data_result.stdout
   args:
     chdir: "{{ openwisp2_path }}"
+
```
pandafy, line 159: @Purhan you should be able to change permissions in the following code. Hence, removing the need for adding a dedicated task for it. 

https://github.com/openwisp/ansible-openwisp2/blob/766a49bdcd06775a9d333dbfea8d3019d58076d1/tasks/django.yml#L3-L8

I believe the idempotence test is also failing due to this. While running above task during idempotence test, molecule observes a change and hence the test suite fails.

pandafy (changes requested):

tasks/django.yml
Purhan: Thanks! I will quickly implement these changes. One question: the issue doesn't mention anything about write permissions, should it be granted to user?

nemesisdesign: I have done the changes I needed locally and polished them. Will send them asap. I suggest to pick another issue, thanks!

Purhan:
Closing after discussion on IM.

PR #247, opened 2020-12-29
[fix] Avoid failure in load_initial_data.py if SSH vars not defined

nemesisdesign:
This can happen when only specific parts of the playbook are run
(eg: by running only specific tags).

### DIFF
--- a/tasks/django.yml
+++ b/tasks/django.yml
@@ -147,8 +147,8 @@
 
 - name: load initial data
   environment:
-    PRIVATE_KEY: "{{ default_private_ssh_key.stdout }}"
-    PUBLIC_KEY: "{{ default_public_ssh_key.stdout }}"
+    PRIVATE_KEY: "{{ default_private_ssh_key.stdout|default(None) }}"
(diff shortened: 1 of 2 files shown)


tasks/django.yml
```diff
@@ -147,8 +147,8 @@
 
 - name: load initial data
   environment:
-    PRIVATE_KEY: "{{ default_private_ssh_key.stdout }}"
-    PUBLIC_KEY: "{{ default_public_ssh_key.stdout }}"
+    PRIVATE_KEY: "{{ default_private_ssh_key.stdout|default(None) }}"
+    PUBLIC_KEY: "{{ default_public_ssh_key.stdout|default(None) }}"
```
pandafy, line 151: Shouldn't these be breezy? Or, is there a specific reason to write it like that? 

```suggestion
    PRIVATE_KEY: "{{ default_private_ssh_key.stdout | default(None) }}"
    PUBLIC_KEY: "{{ default_public_ssh_key.stdout | default(None) }}"
```

pandafy (dismissed):
LGTM :+1: 

I just have a small query.

PR #249, opened 2021-01-05
[fix] Fixed bug affecting installation of custom packages

nemesisdesign:
The variable `openwisp2_notifications_pip` was undefined and invalidating the `when` condition, which caused the list to not be set at all.


pandafy:
I added the variable in documentation in #215 but not in the the code. 🤦

PR #254, opened 2021-01-13
[fix/!change] Reconfigured celery queues to fix SynchronousOnlyOperation exceptions

### LINKED ISSUE #242: [bug/change] Fix SynchronousOnlyOperation: change worker settings
Lately I have started to see many `SynchronousOnlyOperation` since django 3.0.

Looks like using the django ORM from greenlets can lead to thread safety issues.

This means we may not be able to use greenlets as a concurrency pool without refactoring the code to make it greenlet safe.

Greenlet concurrency would have been useful to increase the speed of the monitoring and network requests, but at the moment we may have to go back to prefork (which is based on python multiprocessing).

I have in mind a possible solution like the following:

- switch from gevent to prefork concurrency pool
- create 2 queues: 
 1. one for normal background tasks, which should be the default
 2. one for network tasks (SSH connections like push updates, firmware upgrades, monitoring checks), could be called`network_operations`
- [configure task routing](https://docs.celeryproject.org/en/stable/userguide/routing.html#manual-routing)
- run at least 1 worker process for each queue
- allow to run more worker processes for each queue by configuration
- allow setting the `--concurrency` option or `--autoscale` options (by default these shouldn't be present)

nemesisdesign:
- switched back from gevent to prefork concurrency pool because of thread safety issues
- create 2 queues:
  1. one for normal background tasks, which should be the default
  2. one for network tasks (SSH connections like push updates, firmware upgrades, monitoring checks), could be callednetwork_operations
- configure task routing according to the previous point, allow custom configurations
- run at least 1 worker process for each queue
- allow to run more worker processes for each queue by configuration
- allow setting the --concurrency option or --autoscale options
- removed the "openwisp2_celery_worker_prefetch_multiplier" variable because
  this setting is now passed to the worker configuration

Fixes #242


nemesisdesign (commented):
I tested this and it's working!

I am merging to speed up, but anyone feel free to leave comments or point out to any issue you may find and I'll fix it.

templates/supervisor/celery_network.j2
```diff
@@ -0,0 +1,11 @@
+[program:celery_network]
+user={{ www_user }}
+directory={{ openwisp2_path }}
+command={{ openwisp2_path }}/env/bin/celery -A openwisp2 worker -l info --queue network -O {{ openwisp2_celery_network_optimization }}{% if openwisp2_celery_network_concurrency %} --concurrency={{ openwisp2_celery_network_concurrency }}{% elif openwisp2_celery_network_autoscale %} --autoscale {{ openwisp2_celery_network_autoscale }}{% endif %}{% if openwisp2_celery_network_prefetch_multiplier %} --prefetch-multiplier {{ openwisp2_celery_network_prefetch_multiplier }}{% endif %} -n network@%%h
```
nemesisdesign, line 4: would be great to know if there's a way to break this long line into multiple lines

templates/supervisor/celery.j2
```diff
@@ -1,7 +1,7 @@
 [program:celery]
 user={{ www_user }}
 directory={{ openwisp2_path }}
-command={{ openwisp2_path }}/env/bin/celery -A openwisp2 worker -l info --pool=gevent --concurrency={{ openwisp2_celery_concurrency }} -Ofair
+command={{ openwisp2_path }}/env/bin/celery -A openwisp2 worker -l info --queue celery -O {{ openwisp2_celery_optimization }}{% if openwisp2_celery_concurrency %} --concurrency={{ openwisp2_celery_concurrency }}{% elif openwisp2_celery_autoscale %} --autoscale {{ openwisp2_celery_autoscale }}{% endif %}{% if openwisp2_celery_prefetch_multiplier %} --prefetch-multiplier {{ openwisp2_celery_prefetch_multiplier }}{% endif %} -n background@%%h
```
nemesisdesign, line 4: same here

nemesisdesign (commented):
I tested this and it's working!

I am merging to speed up, but anyone feel free to leave comments or point out to any issue you may find and I'll fix it.

PR #259, opened 2021-02-09
Update pip.yml

marsimiklos:


devkapilbansal:
@marsimiklos why you upgrade pip here. It is not necessary and the pip is pinned to the version to avoid dependency resolution.

PR #262, opened 2021-03-21
[feature] Add support for openwisp-monitoring #195

### LINKED ISSUE #195: [feature] Add openwisp-monitoring
Blocked by https://github.com/openwisp/ansible-openwisp2/issues/214.

- add openwisp-monitoring
  - add celery beat
  - add influxdb, would be great to allow changing the default password easily by just setting it in the playbook
  - add celery task routes for the network queue (related to #242):

```
'openwisp_monitoring.check.tasks.run_checks': {'queue': 'network'},
'openwisp_monitoring.check.tasks.perform_check': {'queue': 'network'},
```

The module will be enabled by default, but it should be possible to disable openwisp monitoring with a variable, like we do for openwisp-network-topology.

Purhan:
Closes #195


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
 
       - name: Install python dependencies
         run: |
-          pip install molecule[docker] yamllint ansible-lint docker openwisp-utils[qa]
+          pip install molecule[docker,ansible] yamllint ansible-lint docker openwisp-utils[qa]
```
pandafy, line 38: I am noot entirely sure that we need this.

defaults/main.yml
```diff
@@ -4,9 +4,11 @@ openwisp2_python: python3
 ansible_python_interpreter: /usr/bin/python3
 openwisp2_network_topology: false
 openwisp2_firmware_upgrader: false
+openwisp2_monitoring: true
```
pandafy, line 7: The default should be `false`

defaults/main.yml
```diff
@@ -16,6 +18,7 @@ openwisp2_django_loci_pip: false
 openwisp2_netjsonconfig_pip: false
 openwisp2_network_topology_pip: false
 openwisp2_firmware_upgrader_pip: false
+openwisp2_monitoring_pip: true
```
pandafy, line 21: The default should be `false`.

@Purhan  we use `openwisp2_**_pip` variables to pass an URL so that pip installs that module from that URL. E.g. if you want to install development version of openwisp-monitoring, you would set

```yml
openwisp2_monitoring_pip: "https://github.com/openwisp/openwisp-monitoring/tarball/master"
```

templates/openwisp2/settings.py
```diff
@@ -53,6 +53,11 @@
 {% if openwisp2_firmware_upgrader %}
     'openwisp_firmware_upgrader',
     'private_storage',
+{% endif %}
+{% if openwisp2_monitoring %}
+    'openwisp_monitoring.monitoring',
+    'openwisp_monitoring.device',
+    'openwisp_monitoring.check',
```
pandafy, line 60: I don't think this will make any much difference, but let's move this block up after `openwisp_controller.connection`. This will make the order consistent with [README of openwisp-monitoring](https://github.com/openwisp/openwisp-monitoring#setup-integrate-in-an-existing-django-project).

pandafy (changes requested):
GG @Purhan 👏🏼 

The following code should start redis by default. I wonder why `migrate` command is failing. 

https://github.com/openwisp/ansible-openwisp2/blob/ee89d2d5ac05004a82e6401a5151297240c02747/tasks/apt.yml#L66-L70

.github/workflows/ci.yml
Purhan: Previously, the build was failing because of this, this was the error I saw:
```
Unable to find a working copy of ansible executable. Read https://molecule.readthedocs.io/en/latest/installation.htmlCompletedProcess(args='ansible --version', returncode=1, stdout='\n', stderr='Traceback (most recent call last):\n  File "/usr/bin/ansible", line 34, in <module>\n    from ansible import context\nModuleNotFoundError: No module named \'ansible\'\n')
```
I saw this error in both my local env and in the github actions CI. Molecule docs [suggested](https://molecule.readthedocs.io/en/latest/installation.html#pip) to add this.

templates/openwisp2/settings.py
```diff
@@ -364,3 +379,14 @@
 {{ instruction }}
 
 {% endfor %}
+
+{% if openwisp2_influxdb_install %}
```
nemesisdesign, line 383: here the condition should be `openwisp2_monitoring`, because even if influxdb is being installed on another server, we still need to configure it

templates/openwisp2/urls.py
```diff
@@ -12,7 +12,7 @@
 
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
-    {% if openwisp2_controller_urls %}
+    {% if openwisp2_controller_urls or openwisp2_monitoring %}
```
nemesisdesign, line 15: why this addition?

defaults/main.yml
nemesisdesign: true is good, most users need this, only advanced users which have a pre-existing monitoring solution which they integrate need to turn this off

nemesisdesign (changes requested):
Great progress @Purhan!

I think we need to add the ansible role which installs influxdb as a dependency: https://github.com/openwisp/ansible-ow-influxdb.

For the rest, please also follow Gagan's comments.

templates/openwisp2/urls.py
Purhan: The [README](https://github.com/openwisp/openwisp-monitoring#setup-integrate-in-an-existing-django-project) of openwisp-monitoring suggests adding `openiwsp-controller`'s urlpatterns. Should I remove this? @nemesisdesign

.github/workflows/ci.yml
nemesisdesign: Ok, thanks for the explanation @Purhan :+1:

Purhan:
I think the build is failing because of an unrelated issue, the tests for openwisp-controller are failing on master when I run them locally

https://github.com/openwisp/ansible-openwisp2/runs/2178256830#step:6:2726

I get the same results on master

templates/openwisp2/urls.py
nemesisdesign: I would remove this change and follow the structure of the current code and add a new var called `openwisp2_monitoring_urls` which allows disabling the URLs if needed.
In some cases the application has to be installed on the VM/container but the URLs may be disabled because are served by another VM/container.

templates/openwisp2/settings.py
```diff
@@ -45,6 +45,11 @@
     'openwisp_controller.config',
     'openwisp_controller.geo',
     'openwisp_controller.connection',
+{% if openwisp2_monitoring %}
+    'openwisp_monitoring.monitoring',
+    'openwisp_monitoring.device',
+    'openwisp_monitoring.check',
+{% endif %}
```
nemesisdesign, line 52: `nested_admin` seems to be missing

nemesisdesign (changes requested):
See my comment below.

A bug seem to affect the build on Ubuntu 18: https://github.com/openwisp/ansible-openwisp2/issues/263

.github/workflows/ci.yml
nemesisdesign, line 38: I had to push this to master since the build was failing without it.

nemesisdesign (changes requested):
Please also rebase on the current master to get rid of the build failure on Ubuntu 18.04 (see #263 and #265 for more info).

nemesisdesign (changes requested):
I updated the README of openwisp-monitoring which was missing nested_admin: https://github.com/openwisp/openwisp-monitoring/commit/bb621ae76e3f4136ef1c8b3d9fdfcab62b960722

From the settings (and from the errors of the build) I see `nested_admin` is missing.
Also while reviewing the instructions of openwisp-monitoring, I see that this is also missing:
```python
INSTALLED_APPS.append('djcelery_email')
EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend' 
```
No need to install `djcelery_email` separately because it's a dependency of ow-monitoring.

templates/openwisp2/settings.py
```diff
@@ -165,12 +170,22 @@
         'schedule': timedelta(days=1),
         'args': ({{ openwisp2_notifications_delete_old_notifications }},),
     },
+    {% if openwisp2_monitoring %}
+    'run_checks': {
+        'task': 'openwisp_monitoring.check.tasks.run_check',
```
nemesisdesign, line 175: it's run_checks, the docs of openwisp-monitoring was incorrect, see https://github.com/openwisp/openwisp-monitoring/issues/281

nemesisdesign (changes requested):

nemesisdesign (commented):
Some of the tests of openwisp-controller are failing because they were not written to work with openwisp-monitoring enabled.

I think we could:

- disable the failing tests if openwisp-monitoring is enabled
- create an issue in openwisp-monitoring to provide some additional integration tests for OpenWISP controller which would inherit the classes that we remove from here and overwrite the tests that are failing
- create an issue here in ansible-openwisp2 to remind us to add back the tests we removed, but, if openwisp-monitoring is enabled, to use the new integration tests provided by openwisp-monitoring rather than the ones of openwisp-controller

@pandafy this is somewhat similar to the issue we had with openwisp-notifications

nemesisdesign (commented):
Is looking good, I think we can undraft this.
I sent a message on the mailing list inviting users to use it and test it.

meta/main.yml
```diff
@@ -8,6 +8,11 @@ dependencies:
     postfix_smtp_sasl_auth_enable: "{{ postfix_smtp_sasl_auth_enable_override | default(false) }}"
     postfix_smtpd_relay_restrictions: "{{ postfix_smtpd_relay_restrictions_override }}"
 
+  - src: https://github.com/purhan/ansible-ow-influxdb
```
nemesisdesign, line 11: I pulled your changes: https://github.com/openwisp/ansible-ow-influxdb/commit/5c61b7e3d7acbe5f0389d7c143bba260253d7b1a
:+1:

molecule/resources/verify.yml
```diff
@@ -5,17 +5,15 @@
   become: true
   gather_facts: false
 
+  vars:
+    openwisp2_monitoring: true
+
```
pandafy, line 10: I don't think we need this since the default value is `true`. Can you confirm @Purhan ?

molecule/resources/verify.yml
```diff
@@ -42,6 +40,16 @@
...
+    - name: Run integration tests when monitoring disabled
+      when: not openwisp2_monitoring
+      command: >
+        /opt/openwisp2/env/bin/python /opt/openwisp2/manage.py test --keepdb \
+            openwisp_controller.config.tests.test_config \
+            openwisp_controller.config.tests.test_controller \
+            openwisp_controller.config.tests.test_device \
+            openwisp_controller.config.tests.test_template \
+            openwisp_controller.config.tests.test_views \
+
```
pandafy, line 52: These tests will always gets skipped while running tests using molecule. Unless we make changes to the test suite and add separate cases where `openwisp2_monitoring` variable is set to false, this is not useful. 

I don't think we need to alter the test suite just to make these tests run since we will be fixing this in near future. What do you think @Purhan ?

pandafy (commented):
Great work @Purhan 👏🏼 

I pulled these changes to deploy an instance of OpenWISP and it worked flawlessly. 

I have shared some queries, can you answer them?

molecule/resources/verify.yml
nemesisdesign: I was thinking that another possible solution is to change those specific tests because the reason they fail is stupid (eg they expect 1 template but there are 2 because openwisp monitoring creates a template, this should be easy to fix in openwisp-controller.

nemesisdesign: Created an issue for it: https://github.com/openwisp/openwisp-controller/issues/442

molecule/resources/verify.yml
Purhan: Yeah right, I will remove it :+1:

molecule/resources/verify.yml
Purhan: So should I remove all these tests for the time being, so that the build passes?

nemesisdesign: I looked into this now. Modifying the tests will take time. It may be more practical to just write some integration tests. So yes, please remove these.

pandafy (changes requested):
@Purhan we also need to update the README for new introduced variables. There is a dedicated section in README for it. See [Role Variables](https://github.com/openwisp/ansible-openwisp2#role-variables)

templates/openwisp2/settings.py
```diff
@@ -368,3 +384,17 @@
...
+    'BACKEND': '{{ openwisp2_timeseries_database.backend }}',
+    'USER': '{{ openwisp2_timeseries_database.user }}',
+    'PASSWORD': '{{ openwisp2_timeseries_database.password }}',
+    'NAME': '{{ openwisp2_timeseries_database.name }}',
+    'HOST': '{{ influxdb_http_ip }}',
+    'PORT': '{{ influxdb_http_port }}',
+}
+
+INSTALLED_APPS.append('djcelery_email')
+EMAIL_BACKEND = 'djcelery_email.backends.CeleryEmailBackend'
```
pandafy, line 399: We should have a separate setting for managing `djcelery_email`. The setting should be `true` by default.

templates/openwisp2/settings.py
```diff
@@ -368,3 +384,17 @@
...
+{% if openwisp2_monitoring %}
+TIMESERIES_DATABASE = {
+    'BACKEND': '{{ openwisp2_timeseries_database.backend }}',
+    'USER': '{{ openwisp2_timeseries_database.user }}',
+    'PASSWORD': '{{ openwisp2_timeseries_database.password }}',
+    'NAME': '{{ openwisp2_timeseries_database.name }}',
+    'HOST': '{{ influxdb_http_ip }}',
+    'PORT': '{{ influxdb_http_port }}',
+}
+
```
pandafy, line 397: @Purhan move this block above https://github.com/openwisp/ansible-openwisp2/pull/262/files#diff-8c46e786c6289dcbdc2b3e43782ff1cd1308bf2cde778ac89485ea48da22095bL362-L368

pandafy (changes requested):

templates/openwisp2/settings.py
nemesisdesign: we can add a variable like `openwisp2_email_backend` which defaults to `djcelery_email.backends.CeleryEmailBackend`

templates/openwisp2/settings.py
nemesisdesign: above what?

pandafy: https://github.com/openwisp/ansible-openwisp2/blob/f05a3b79e537482a513aa0ac40914d22c4a2f228/templates/openwisp2/settings.py#L378-L386

templates/openwisp2/settings.py
Purhan: Done (https://github.com/openwisp/ansible-openwisp2/pull/262/commits/59796a1e231670bffeb084e5a890fda4fc64a3b8)

templates/openwisp2/settings.py
Purhan: Don (https://github.com/openwisp/ansible-openwisp2/pull/262/commits/6d01973916f78d9f92763b17e77f2bf9c2bb5509) :+1:

templates/openwisp2/settings.py
```diff
@@ -359,6 +375,20 @@
...
+INSTALLED_APPS.append('djcelery_email')
+EMAIL_BACKEND = '{{ openwisp2_email_backend }}'
```
pandafy, line 389: It should not be in this block @Purhan. Since we added a dedicated setting to allow configuring email settings, it should be universal. Users should be able to configure it without using openwisp-monitoring. 

Do you understand what I mean here?

templates/openwisp2/settings.py
```diff
@@ -359,6 +375,20 @@
...
+TIMESERIES_DATABASE = {
+    'BACKEND': '{{ openwisp2_timeseries_database.backend }}',
+    'USER': '{{ openwisp2_timeseries_database.user }}',
+    'PASSWORD': '{{ openwisp2_timeseries_database.password }}',
+    'NAME': '{{ openwisp2_timeseries_database.name }}',
+    'HOST': '{{ influxdb_http_ip }}',
+    'PORT': '{{ influxdb_http_port }}',
+}
+
+INSTALLED_APPS.append('djcelery_email')
```
pandafy, line 388: I still think we should add a dedicated setting for enable `djcelery_email` @nemesisdesign 
We can use that to install `djcelery_email` and add needed settings. 
It will also allow us to move ahead with https://github.com/openwisp/openwisp-monitoring/pull/254

It will also open possibility to add a dedicated queue for emails. What do you think about this?

nemesisdesign: @pandafy what is the use case for that? Why should we do this additional effort, what benefit do we get back from it?
I feel that this form is enough to do what we and most users need and if anyone needs to override this they can do it, but I don't see a good reason to change this part. Sending emails in the background is always a good idea.

pandafy: If we set `openwisp2_email_backend` to use djcelery_email backend as default and do not install the package by default, that will make users run into errors. 

We should install that package by default and not just limit it to be used when openwisp-monitoring is enabled.

nemesisdesign: ok @pandafy but what you ask is a different kind of change (make django-celery-email the default email backend) and not strictly related to openwisp-monitoring? I think how it was before was good enough for introducing openwisp-monitoring to ansible-openwisp2, which is the goal we're interested in this PR.

I suggest simply reverting https://github.com/openwisp/ansible-openwisp2/pull/262/commits/59796a1e231670bffeb084e5a890fda4fc64a3b8 and create a new issue to make the django-celery-email backend the default one.

nemesisdesign (changes requested):
@Purhan can you update the README as noted by @pandafy?

templates/openwisp2/settings.py
pandafy: ok @nemesisdesign, opened https://github.com/openwisp/ansible-openwisp2/issues/271

README.md
```diff
@@ -585,6 +585,8 @@ Below are listed all the variables you can customize (you may also want to take
     openwisp2_network_topology_version: "0.4"
     openwisp2_firmware_upgrader: false
     openwisp2_firmware_upgrader_version: "0.1"
+    openwisp2_monitoring: true
```
nemesisdesign, line 588: Reminder for me: I have to update the README to add some information about the monitoring module.

templates/openwisp2/settings.py
```diff
@@ -359,6 +374,20 @@
...
 {% endif %}
 
+{% if openwisp2_monitoring %}
+TIMESERIES_DATABASE = {
+    'BACKEND': '{{ openwisp2_timeseries_database.backend }}',
+    'USER': '{{ openwisp2_timeseries_database.user }}',
+    'PASSWORD': '{{ openwisp2_timeseries_database.password }}',
+    'NAME': '{{ openwisp2_timeseries_database.name }}',
+    'HOST': '{{ influxdb_http_ip }}',
+    'PORT': '{{ influxdb_http_port }}',
```
pandafy, line 384:
```suggestion
    'HOST': '{{ openwisp2_timeseries_database.host }}',
    'PORT': '{{ openwisp2_timeseries_database.port }}',
```

nemesisdesign:
I have merged this manually in the `dev` branch, I will keep this open until `dev` is ready to be recommended as a way to install OpenWISP Monioring, something I'm working on with @pandafy.

nemesisdesign:
Merged manually in the `dev` branch, this can be closed, thank you very much @purhan @pandafy! :+1: :+1:

PR #268, opened 2021-04-29
Document openwisp2_django_loci_pip variable in readme.

pniaps:
I've added `openwisp2_django_loci_pip` to README.md and also moved `openwisp2_notifications_pip` after `openwisp2_controller_pip` to respect the order in which they are defined in `defaults/main.yml` file


nemesisdesign (approved):
Thanks

PR #270, opened 2021-05-06
Remove deprecated django.core.urlresolvers

pniaps:
django.core.urlresolvers was deprecated in Django 1.10 and removed in Django 2.0
openwisp-controller requires at least Django 3.0


nemesisdesign (approved):
Thanks!

PR #282, opened 2021-07-29
[change] Use UNIX sockets for uwsgi and daphne #281

### LINKED ISSUE #281: [change] Change nginx conf to use sockets to redirect traffic to uwsgi and daphne
Since the traffic is on the same host, we can use unix sockets to reduce overhead.

For daphne:
```
upstream channels-backend {
  server unix:{openwisp2_path}/daphne0.sock fail_timeout=0;
}
```

For uwsgi: https://uwsgi-docs.readthedocs.io/en/latest/Nginx.html#configuring-nginx and I think we'll have to configure the uwsgi.ini to use the unix socket instead of the TCP socket.

pandafy:
Closes #281


templates/supervisor/daphne.j2
```diff
@@ -1,6 +1,6 @@
 [fcgi-program:daphne]
 user={{ www_user }}
-socket=tcp://localhost:8000
+socket=unix://{{ openwisp2_path }}/daphne.sock
```
nemesisdesign, line 3: this points to `daphne.sock`

templates/nginx.j2
```diff
@@ -1,5 +1,5 @@
 upstream channels-backend {
-    server localhost:8000;
+    server unix://{{ openwisp2_path }}/daphne0.sock fail_timeout=0;
```
nemesisdesign, line 2: this one points to `daphne0.sock`, are you sure this difference is not a problem?

nemesisdesign (approved):

PR #286, opened 2021-08-25
[fix] correct task and defaults related to `openwisp2_monitoring` changes

broadstack-au:
Hi,

(related to changes introduced in #195 #262)

I was testing out the role so we could add openwisp-monitoring into our ansible deployed controller.
These two files were triggering errors, so I figured I'd push the fixes back.

* `defaults/main.yml` provides defaults for all bar host and port. Given the linked role installs influxdb appropriately without those settings - it seemed sensible to just put the defaults in.

* `tasks/apt.yaml` had old/invalid syntax. I've just updated the task to be the same as previous apt tasks.

This was all we needed to modify to deploy to our pre-existing instance.

Do you folk need any assistance or further work for the other PR? We'd really like to see it make the next release :)

Thanks!

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -135,8 +135,7 @@
 
 - name: Install fping
   when: openwisp2_monitoring is true
-  apt:
-    - fping
+  apt: name=fping
   retries: 5
   delay: 10
   register: result
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -96,6 +96,8 @@ openwisp2_timeseries_database:
     user: "openwisp"
     password: "openwisp"
     name: "openwisp2"
+    host: "localhost"
+    port: 8086
 openwisp2_nginx_access_log: "{{ openwisp2_path }}/log/nginx.access.log"
 openwisp2_nginx_error_log: "{{ openwisp2_path }}/log/nginx.error.log error"
 # keep backward compatibility with old variable openwisp2_eventlet_concurrency


tasks/apt.yml
```diff
@@ -135,8 +135,7 @@
 
 - name: Install fping
   when: openwisp2_monitoring is true
-  apt:
-    - fping
+  apt: name=fping
```
nemesisdesign, line 138: could you share info about changes from old syntax to the new one?
We have quite a lot of this so if it still works we could avoid changing it now and change it all at once later on.

nemesisdesign (commented):
> Hi,
> 
> (related to changes introduced in #195 #262)
> 
> I was testing out the role so we could add openwisp-monitoring into our ansible deployed controller.
> These two files were triggering errors, so I figured I'd push the fixes back.

Thank you for taking the time to do this.

> * `defaults/main.yml` provides defaults for all bar host and port. Given the linked role installs influxdb appropriately without those settings - it seemed sensible to just put the defaults in.
> * `tasks/apt.yaml` had old/invalid syntax. I've just updated the task to be the same as previous apt tasks.
> 
> This was all we needed to modify to deploy to our pre-existing instance.
> 
> Do you folk need any assistance or further work for the other PR? We'd really like to see it make the next release :)

We have been busy on different fronts introducing many improvements and features, deploying and using all this work, so the way you can help us is to keep doing what you've been doing (deploy, test, fix) and let us know if you find any issue, if more people did that we'd surely be able to advance faster, thanks.

pandafy:
Thank you for your contribution @broadstack-au! I have manually added your changes in https://github.com/openwisp/ansible-openwisp2/pull/262/commits/9bf1186f89dd203a5cb0d9c6881d361e0eb718e5 after making some changes.

PR #287, opened 2021-08-25
[fix] Honour variables for nginx logrotate config

broadstack-au:
The role allows the `openwisp2_path` variable to be modified, however the nginx logrotate config was a plain file with hardcoded paths and user/group.

This change updates the task to use `template` instead of `copy` and honours the `openwisp2_path`, `www_user` & `www_group` variables defined by the play.

No material changes were observed during testing - logrotate functions as it did before.

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -46,6 +46,7 @@
   notify: restart nginx
 
 - name: configure nginx log rotation
-  copy:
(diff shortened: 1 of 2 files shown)


pandafy (approved):
Thank you @broadstack-au! LGTM 👍🏼

PR #296, opened 2021-09-23
[feature] Added support for internationalization #290

### LINKED ISSUE #290: [feature] Add support for internationalization
The steps required to make internationalization work are the following ones:

Add `'django.middleware.locale.LocaleMiddleware'` to `MIDDLEWARE` in `settings.py` (before CommonMiddleware).

Install gettext, eg: `apt install gettext`

Run `compilemessages` (could be done after `collectstatic`):

```
./manage.py compilemessages
```

Ensure the following settings are `True`:

```python
USE_I18N = True
USE_L10N = True
```

We could add a variable which controls whether this is enabled or not, it may be disabled by default, because if internationalization is not required, disabling it makes the system faster.

codesankalp:
To support internationalization in openwisp2 added a role valriable
openwisp2_internationalization which is by default false but when enabled
compiles messages.

Closes #290


templates/openwisp2/settings.py
```diff
@@ -102,6 +102,7 @@
 MIDDLEWARE = [
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.locale.LocaleMiddleware',
```
nemesisdesign, line 105: this should also depend on openwisp2_internationalization

tasks/django.yml
```diff
@@ -139,6 +139,27 @@
...
+  file:
+    path: "{{ virtualenv_path }}"
+    state: directory
+    group: "{{ www_group }}"
+    owner: "{{ www_user }}"
+    # sets executable only to directories
+    mode: u=rwX,g=rX,o=rX
+    recurse: true
+  tags:
+    - molecule-idempotence-notest
```
nemesisdesign, line 152: is this really needed?

tasks/django.yml
```diff
@@ -139,6 +139,27 @@
...
+
+- name: compilemessages
+  notify: reload supervisor
+  become: true
+  become_user: "{{ www_user }}"
+  django_manage:
+    app_path: "{{openwisp2_path}}"
+    command: "compilemessages"
+    virtualenv: "{{ virtualenv_path }}"
```
nemesisdesign, line 161: this should depend on `openwisp2_internationalization`, using `when: openwisp2_internationalization`.

nemesisdesign (changes requested):

tasks/django.yml
codesankalp: Yes, without this error is shown by ansible: https://github.com/django/django/blob/25cbd1e6aa4178c31d60e3900a5bd4aa177c01c9/django/core/management/commands/compilemessages.py#L149-L155

But on running manually it is not showing errors.

nemesisdesign: Ok, please add this reason as a comment preceding the task

Issue #285, opened 2021-08-22
[feature] Add support for Debian 11 (Bullseye)


t-trmg:
Hello,

I updated my OpenWISP controller to Debian 11 (Bullseye) which has broken things.  This I was completely expecting.  So, I attempted to re-install via the upgrade instructions and the following issues occur:

1) The step where redis is installed, it fails.  I believe this is because the current version of redis packaged for Debian 11 is 6.0.15 whereas the Ansible installer is looking for either 5.5 or 5.4.

```
TASK [openwisp.openwisp2 : Install redis 5 if available] *************************************************************
fatal: [openwisp.vofr.net]: FAILED! => {"cache_update_time": 1629652773, "cache_updated": false, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'redis-server=5:5*'' failed: E: Version '5:5*' for 'redis-server' was not found\n", "rc": 100, "stderr": "E: Version '5:5*' for 'redis-server' was not found\n", "stderr_lines": ["E: Version '5:5*' for 'redis-server' was not found"], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information..."]}
...ignoring

TASK [openwisp.openwisp2 : Install redis 4 (fallback)] ***************************************************************
fatal: [openwisp.vofr.net]: FAILED! => {"cache_update_time": 1629652773, "cache_updated": false, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'redis-server=5:4*'' failed: E: Version '5:4*' for 'redis-server' was not found\n", "rc": 100, "stderr": "E: Version '5:4*' for 'redis-server' was not found\n", "stderr_lines": ["E: Version '5:4*' for 'redis-server' was not found"], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information..."]}
...ignoring
```

2) The update pip & related tools step fails as well.  This one I'm unsure as to why.  The current version of python3 appears to be 3.9.2 if this helps.

```
TASK [openwisp.openwisp2 : Update pip & related tools] ***************************************************************
[...]

PR #299, opened 2021-09-27
[feature] Debian Bullseye/redis 6 support

peschmae:
Debian Bullseye ships with redis 6, so I've added a task that installs redis 6.

Additionally I´ve had to adjust the check in `tasks/pip.yml` to make use of the newly registred variable.

The molecule tests (locally, as well as in github actions), are now also run against debian11 (again using the images by geerlingguy)

Improves on #285 
I wasn't able to reproduce the pip error mentioned in the ticket, so I didn´t change anything on these tasks

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -53,14 +53,13 @@
   when: openwisp2_redis_install and use_redis6 is failed
 
 - name: Install redis 4 (fallback)
-  when: openwisp2_redis_install and (use_redis5 is failed or use_redis6 is failed)
+  when: openwisp2_redis_install and (use_redis5 is failed and use_redis6 is failed)
   apt:
     name: "redis-server=5:4*"
   register: use_redis4
   notify:
     - reload systemd
     - start redis
-  ignore_errors: true
 
 # On the newer versions of redis, by default redis
 # binds to localhost on ipv6 address which wouldn't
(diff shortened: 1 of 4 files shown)


tasks/apt.yml
```diff
@@ -40,12 +50,13 @@
     - start redis
   ignore_errors: true
   register: use_redis5
-  when: openwisp2_redis_install
+  when: openwisp2_redis_install and use_redis6 is failed
 
 - name: Install redis 4 (fallback)
-  when: openwisp2_redis_install and use_redis5.failed
+  when: openwisp2_redis_install and (use_redis5 is failed or use_redis6 is failed)
```
pandafy, line 56: In the current state this task will still execute if use_redis6 failed but use_redis5 was successful. 

It has happened in the build

https://github.com/openwisp/ansible-openwisp2/pull/299/checks?check_run_id=3723105424#step:7:270

It should be any one from below. 

```suggestion
  when: openwisp2_redis_install and (use_redis5 is failed and use_redis6 is failed)
```


use_redis5 will only get registered with redis6 failed. 
```suggestion
  when: openwisp2_redis_install and use_redis5 is failed
```

Also, this task should not ignore error.

pandafy (changes requested):

pandafy (approved):

nemesisdesign (approved):

nemesisdesign:
Many thanks @peschmae :+1:

PR #315, opened 2021-11-12
[change] Removed sqlcounters by default #314

### LINKED ISSUE #314: [change:dev-branch] Remove sqlcounters by default
Once https://github.com/openwisp/openwisp-radius/pull/334 is merged, a basic mutli-tenant compatible sqlcounter implementation will be shipped in OpenWISP RADIUS.

For this reason, we should avoid enabling by default the sql counters in the [default freeradius site conf](https://github.com/openwisp/ansible-openwisp2/blob/dev/templates/freeradius/openwisp_site.j2), otherwise multiple queries would be repeated twice, generating usless additional load on the resources of the system.

We should also remove these templates from `templates/freeradius` and any task which reference them:

- `daily-bandwidth-counter`,
- `daily-counter`
- `monthly-counter`

I think we can then also remove the template `templates/freeradius/sql_counter.j2` and the task which uploads it, since it's a modified version of the default `sqlcounter` module.

I think most of the tasks mentioning the word "counter" here can be removed:
https://github.com/openwisp/ansible-openwisp2/blob/f71187fc4895abc8c4f06b5268a527b7bf0583d2/tasks/freeradius.yml#L115-L122

After this changes it's paramount to fully run the installation process on a test VM or docker container and ensure [that freeradius can start properly and a basic authorization with radtest works](https://openwisp-radius.readthedocs.io/en/latest/developer/freeradius.html#testing-authentication-and-authorization).

codesankalp:
Closes #314


nemesisdesign (changes requested):
@codesankalp I think this branch may need to be rebased on the latest `dev` branch.

nemesisdesign (changes requested):
@codesankalp why is 59bb635e4413808d76b0b408cc7eb7e4255abf46 included in this PR? Is that intentional? It doesn't look like it.

files/storage.py
```diff
@@ -7,6 +7,21 @@
 
 
 class FileHashedNameMixin:
+    # Workaround for modules that do not completely support Django 4.
+    # TODO: Remove this while adding Django 4.0 support.
```
nemesisdesign, line 11: I don't understand this TODO comment very well, we should remove this while adding Django 4.0 support? What does that mean? Can you clarify please?

nemesisdesign (changes requested):
I think the last two commits in this PR belong to their own PR, please open a new PR just with those two.

PR #316, opened 2021-11-16
[change] Added ansible variable to set max file size for upload #269

### LINKED ISSUE #269: [change] Update nginx max file upload size to match firmware upgrader
If firmware upgrader is enabled, we should increase the default nginx max file upload size to 30 MB to match https://github.com/openwisp/openwisp-firmware-upgrader#openwisp-firmware-upgrader-max-file-size.

I think the easiest thing may be to add a way to control `OPENWISP_FIRMWARE_UPGRADER_MAX_FILE_SIZE` with an ansible variable, and use the same variable also to configure the nginx max file upload size.

pandafy:
"openwisp2_firmware_upgrader_max_file_size" sets
"OPENWISP_FIRMWARE_UPGRADER_MAX_FILE_SIZE" in settings.py and
 updates "client_max_body_size" in nginx config.

Closes #269

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -1,4 +1,9 @@
 ---
+- name: Set nginx max upload size
+  # Required to allow uploading firmware images
+  set_fact:
+    openwisp2_nginx_client_max_body_size: "{{openwisp2_firmware_upgrader_max_file_size}}"
+  when: openwisp2_firmware_upgrader
 
 - name: create "{{ openwisp2_path }}/public_html"
   file:
--- a/README.md
+++ b/README.md
@@ -1005,6 +1005,10 @@ Below are listed all the variables you can customize (you may also want to take 
     postfix_smtpd_relay_restrictions_override: permit_mynetworks
     # allows overriding the default duration for keeping notifications
     openwisp2_notifications_delete_old_notifications: 10
+    # Maximum file size(in bytes) allowed to be uploaded as firmware image.
+    # It overrides "openwisp2_nginx_client_max_body_size" setting
+    # and updates nginx configuration accordingly.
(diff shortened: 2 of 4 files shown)


tasks/nginx.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: Set nginx max upload size
+  # Required to allow uploading firmware images
+  set_fact:
+    openwisp2_nginx_client_max_body_size: "{{openwisp2_firmware_upgrader_max_file_size}}"
+  when: openwisp2_firmware_upgrader
```
nemesisdesign, line 6: I would add `and not openwisp2_nginx_client_max_body_size` so that we may be able to set a higher size to `openwisp2_nginx_client_max_body_size` if ever needed (imagine a different module which requires even bigger upload sizes). What do you think?

nemesisdesign (commented):
All good, I have one suggestion below.

tasks/nginx.yml
pandafy: `openwisp2_nginx_client_max_body_size` defaults to `20M`. We will have to change few other things to make it work. But, it is doable.

nemesisdesign: ah ok sorry I forgot, no problem, let's keep it simple then

nemesisdesign: here I have an idea, the condition could be:

```
when: openwisp2_firmware_upgrader and openwisp2_firmware_upgrader > openwisp2_nginx_client_max_body_size
```

What do you think?

pandafy: `openwisp2_firmware_upgrader_max_file_size` is bytes in `int`. `openwisp2_nginx_client_max_body_size` can be that, or in format like `10K`/``20M`. It won't work in all cases.

PR #321, opened 2021-12-06
[change] Added variable to configure daphne websocket timeout

pandafy:
This timeout value is also used for configuring the "group_expiry"
of CHANNEL_LAYERS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -910,6 +910,8 @@ Below are listed all the variables you can customize (you may also want to take 
     openwisp2_uwsgi_threads: 2
     # number of daphne process to spawn. Default value is 1
     openwisp2_daphne_processes: 2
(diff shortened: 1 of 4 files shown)


nemesisdesign (commented):
Let's check whether the frontend JS code can handle the timeout and reconnect.

meta/main.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
     postfix_smtpd_relay_restrictions: "{{ postfix_smtpd_relay_restrictions_override }}"
 
   - src: https://github.com/openwisp/ansible-ow-influxdb
-    version: origin/master
+    version: origin/issues/3-updating-user-password
```
nemesisdesign, line 12: looks like this will have to be updated

nemesisdesign (approved):

PR #323, opened 2021-12-13
[change] Added option to upload custom static files #322

### LINKED ISSUE #322: [feature] Add STATICFILES_DIR and possibility to easily upload custom static files
Now that https://github.com/openwisp/openwisp-utils/pull/266 is merged, it became evident that Django needs to be configured properly in order for the staticfiles framework to find the custom static files.

We need to do this in a systematic way and document it.

For example, we could designate `static/custom/` as the directory where to place custom static files, then add to `settings.py`:

```python
STATICFILES_DIRS = [os.path.join(BASE_DIR, 'static/custom')]
```

Then we can make it easy to upload files to that location in a similar way as we are doing in for [ansible-openwisp-wifi-login-pages](https://github.com/openwisp/ansible-openwisp-wifi-login-pages#deploying-custom-static-content).

pandafy:
Closes #322

### DIFF
--- a/templates/openwisp2/settings.py
+++ b/templates/openwisp2/settings.py
@@ -335,6 +335,7 @@ USE_TZ = True
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/1.9/howto/static-files/
 
+STATICFILES_DIRS = [os.path.join(BASE_DIR, 'openwisp2/static/custom')]
 STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 STATIC_URL = '/static/'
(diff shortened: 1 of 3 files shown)


templates/openwisp2/settings.py
```diff
@@ -335,6 +335,7 @@
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/1.9/howto/static-files/
 
+STATICFILES_DIRS = [os.path.join(BASE_DIR, 'openwisp2/static/custom')]
```
nemesisdesign, line 338: shouldn't be this put one line below and be:

```python
STATICFILES_DIRS = [os.path.join(STATIC_ROOT, 'custom')]
```

nemesisdesign (changes requested):

templates/openwisp2/settings.py
nemesisdesign: Maybe we can do this:
```
/static
/static-custom
```

PR #330, opened 2022-01-10
[fix] Fixed support for Ubuntu 18.04

pandafy:
- Role will install Python 3.7 if Python version < 3.7 is found.
- Pinned setuptools~=59.6.0


nemesisdesign (approved):
Thanks :+1:

PR #336, opened 2022-01-20
[feature] Added support for openwisp_controller.subnet_division #334

### LINKED ISSUE #334: [feature] Add support for automatic provisioning of subnets
The new version of openwisp-controller will bring the feature that implements automatic provisioning of subnets. 
https://github.com/openwisp/openwisp-controller/tree/1.0.x#subnet-division-app

Since this feature might not be required by everyone, we should add an ansible variable through which we can enable/disable the feature. 

By default, the feature should be disabled.

pandafy:
- openwisp-ipam will be installed and enabled by default.

Closes #334

**Blockers**

- https://github.com/openwisp/openwisp-controller/pull/591

### DIFF
--- a/README.md
+++ b/README.md
@@ -796,6 +796,11 @@ Below are listed all the variables you can customize (you may also want to take 
     openwisp2_network_topology: false
     openwisp2_firmware_upgrader: false
     openwisp2_radius: false
+    # Setting this to true will enable subnet division feature of
+    # openwisp-controller. Refer openwisp-controller documentation
+    # for more information. https://github.com/openwisp/openwisp-controller#subnet-division-app
+    # By default, it is set to false.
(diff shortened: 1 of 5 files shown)


tasks/pip.yml
```diff
@@ -83,7 +83,7 @@
   pip:
     name:
       # - "openwisp-controller~={{ openwisp2_controller_version }}"
-      - "https://github.com/openwisp/openwisp-controller/tarball/master"
+      - "https://github.com/openwisp/openwisp-controller/tarball/decouple-subnet-division"
```
pandafy, line 86: TODO: Remove this before merging.

nemesisdesign (approved):
:+1:

PR #350, opened 2022-02-28
[enhancement] Allow users to specify Django version #313

### LINKED ISSUE #313: [enhancement] Allow users to specify Django version
Add a variable to specify which version of Django should be installed. 
This task should be run after everything is installed by pip to avoid accidental upgrades by any other package. 

I think we should default it to `~=3.2.0`, since it is an LTS release of Django (open for discussion).

yashikajotwani12:
Fixes #313

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -13,7 +13,7 @@ openwisp2_network_topology_version: "0.5.1"
 openwisp2_firmware_upgrader_version: "0.1"
 openwisp2_monitoring_version: "0.1"
 openwisp2_radius_version: "0.2.1"
-openwisp2_django_version: "3.2.0"
+openwisp2_django_version: "~=3.2.0"
 openwisp2_controller_pip: false
 openwisp2_notifications_pip: false
 openwisp2_users_pip: false
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -13,6 +13,7 @@ openwisp2_network_topology_version: "0.5.1"
 openwisp2_firmware_upgrader_version: "0.1"
 openwisp2_monitoring_version: "0.1"
 openwisp2_radius_version: "0.2.1"
+openwisp2_django_version: "3.2.0"
 openwisp2_controller_pip: false
 openwisp2_notifications_pip: false
 openwisp2_users_pip: false
--- a/tasks/pip.yml
+++ b/tasks/pip.yml
@@ -347,7 +347,7 @@
 
 - name: Install django
   pip:
-    name: "django~=3.2.0"
+    name: "django{{ openwisp2_django_version}}"
     state: present
     virtualenv: "{{ virtualenv_path }}"
     virtualenv_python: "{{ openwisp2_python }}"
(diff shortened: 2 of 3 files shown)


tasks/pip.yml
```diff
@@ -328,3 +328,18 @@
   register: result
   until: result is success
   notify: reload supervisor
+
+- name: Install django
+  pip:
+    name: "django~=3.2.0"
```
pandafy, line 334: @yashikajotwani12  we need to add a variable that contains `~=3.2.0`. 

```suggestion
    name: "django{{ openwisp2_django_version}}"
```

This variable should be added in https://github.com/openwisp/ansible-openwisp2/blob/master/defaults/main.yml with the default value of `~=3.2.0`. 

Please also update the ["Role Variables" section of the README](https://github.com/openwisp/ansible-openwisp2#role-variables) to add this new variable.

pandafy (changes requested):
Good start @yashikajotwani12! See my comment below.

tasks/pip.yml
yashikajotwani12: Yeah! I will update these changes soon, Thanks

defaults/main.yml
```diff
@@ -13,6 +13,7 @@ openwisp2_network_topology_version: "0.5.1"
 openwisp2_firmware_upgrader_version: "0.1"
 openwisp2_monitoring_version: "0.1"
 openwisp2_radius_version: "0.2.1"
+openwisp2_django_version: "3.2.0"
```
pandafy, line 16:
```suggestion
openwisp2_django_version: "~=3.2.0"
```
It won't work in the current state.

tasks/pip.yml
```diff
@@ -328,3 +328,18 @@
   register: result
   until: result is success
   notify: reload supervisor
+
+- name: Install django
+  pip:
+    name: "django{{ openwisp2_django_version}}"
```
pandafy, line 334: In the current state, this line will evaluate to `django3.2.0` which is wrong.

pandafy (changes requested):

tasks/pip.yml
yashikajotwani12: Okay, I understood. I am improving this. Thanks

nemesisdesign:
```suggestion
    name: "django~={{ openwisp2_django_version}}"
```

defaults/main.yml
```diff
@@ -13,6 +13,7 @@ openwisp2_network_topology_version: "0.5.1"
 openwisp2_firmware_upgrader_version: "0.1"
 openwisp2_monitoring_version: "0.1"
 openwisp2_radius_version: "0.2.1"
+openwisp2_django_version: "~=3.2.0"
```
nemesisdesign, line 16: This cant' be so inconsistent with the rest of the code, moreover, the README doesn't reflect this, too many wrongs here.

```suggestion
openwisp2_django_version: "3.2.0"
```

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #371, opened 2022-04-29
[fix] Install freeradius from network radius packages #368

### LINKED ISSUE #368: [bug] Install freeradius from network radius packages
The freeradius package shipped right now in many systems does not work with HTTP2.

We should install the [freeradius packages provided by network radius](https://networkradius.com/packages/) which are more up to date.

codesankalp:
Fixes #368

### DIFF
--- a/tasks/freeradius.yml
+++ b/tasks/freeradius.yml
@@ -1,8 +1,91 @@
 ---
+- name: Add freeradius repository into Ubuntu focal sources list
+  shell: |
+    echo "deb http://packages.networkradius.com/freeradius-3.0/ubuntu/focal focal main" | \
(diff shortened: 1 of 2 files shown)


tasks/freeradius.yml
```diff
@@ -1,8 +1,17 @@
 ---
+- name: Add freeradius repository into sources list
+  ansible.builtin.apt_repository:
+    repo: deb http://packages.networkradius.com/freeradius-3.0/debian/bullseye bullseye main
```
codesankalp, line 4: In this do we have to check the ansible os family first and then add the apt repository?

nemesisdesign: correct

nemesisdesign (approved):

PR #374, opened 2022-05-04
[chores] Added task to trigger timeseries database migration 

pandafy:
Related to https://github.com/openwisp/openwisp-monitoring/pull/368

### DIFF
--- a/templates/openwisp2/settings.py
+++ b/templates/openwisp2/settings.py
@@ -280,7 +280,7 @@ CELERY_TASK_ROUTES = {
 {% if openwisp2_monitoring and openwisp2_celery_monitoring %}
     # monitoring checks are executed in a dedicated "monitoring" queue
     'openwisp_monitoring.check.tasks.perform_check': {'queue': 'monitoring'},
-    'openwisp_monitoring.monitoring.tasks.migrate_timeseries_database': {'queue': 'influxdb_migration'},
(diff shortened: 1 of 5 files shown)


nemesisdesign (changes requested):
Looks good,!

In order to make sure the migration runs smoothly on most instances, I suggest we do the following:

- set the default of `openwisp2_celery_monitoring_concurrency` to 2
- make sure this task is routed to the monitoring worker

This way we ensure the task doesn't occupy a worker on default settings, which may halt the background operations of other tasks.

templates/openwisp2/settings.py
```diff
@@ -280,6 +280,7 @@
 {% if openwisp2_monitoring and openwisp2_celery_monitoring %}
     # monitoring checks are executed in a dedicated "monitoring" queue
     'openwisp_monitoring.check.tasks.perform_check': {'queue': 'monitoring'},
+    'openwisp_monitoring.monitoring.tasks.migrate_timeseries_database': {'queue': 'influxdb_migration'},
```
nemesisdesign, line 283: I think this should be "monitoring", we don't have a queue called `influxdb_migration` here, must be a leftover.

nemesisdesign (changes requested):
I found one problem below, the rest looks good and once the [last doubt on the openwisp-monitoring PR](https://github.com/openwisp/openwisp-monitoring/pull/368#pullrequestreview-962741150) is resolved we can update this PR to point to master again and merge.

PR #380, opened 2022-05-08
[feature] Added support for django CORS #69

### LINKED ISSUE #69: [django] Add support for CORS (Cross-Origin Resource Sharing)
We need to make it possible to enable CORS because [other applications may need to access some APIs to get data](https://lists.funkfeuer.at/pipermail/interop-dev/2017-December/000732.html).

We can use this module: [django-cors-headers](https://pypi.python.org/pypi/django-cors-headers/2.1.0) - which is configurable via django settings, we have to make sure we can configure every available setting of this module, possibly using a single ansible variable containing a dictionary.
  
This feature should be optional and turned off by default now.

Aryamanz29:
-  Added support for `django-cors-headers`.
-  Updated `README`.

Closes #69

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -177,3 +177,22 @@ cron_delete_old_users: "'hour': 0, 'minute': 10"
 cron_cleanup_stale_radacct: "'hour': 0, 'minute': 20"
 cron_delete_old_postauth: "'hour': 0, 'minute': 30"
 cron_delete_old_radacct: "'hour': 1, 'minute': 30"
+
+# cross-origin resource sharing (CORS) settings
+django_cors:
+    enabled: false
+    allow_all_origins: false
+    allowed_origins_list: []
+    allowed_origins_regexes: false
+    allowed_origins_regexes_list: []
+    urls_regex: r'^.*$'
+    allow_methods_default: true
+    custom_methods_list: []
+    allow_headers_default: true
+    custom_headers_list: []
+    expose_headers_list: []
+    preflight_max_age: 86400
+    allow_credentials: false
+    # CSRF Integration
+    csrf_trusted_origin: []
+    replace_https_referer: false
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -177,22 +177,8 @@ cron_delete_old_users: "'hour': 0, 'minute': 10"
 cron_cleanup_stale_radacct: "'hour': 0, 'minute': 20"
 cron_delete_old_postauth: "'hour': 0, 'minute': 30"
 cron_delete_old_radacct: "'hour': 1, 'minute': 30"
-
-# cross-origin resource sharing (CORS) settings
-django_cors:
+# Cross-Origin Resource Sharing (CORS) settings
+openwisp2_django_cors:
     enabled: false
-    allow_all_origins: false
     allowed_origins_list: []
-    allowed_origins_regexes: false
-    allowed_origins_regexes_list: []
-    urls_regex: r'^.*$'
-    allow_methods_default: true
-    custom_methods_list: []
-    allow_headers_default: true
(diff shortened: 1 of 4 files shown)


templates/openwisp2/settings.py
```diff
@@ -494,3 +506,29 @@
 {{ instruction }}
 
 {% endfor %}
+
+{% if django_cors.enabled %}
+
```
nemesisdesign, line 511: to avoid scattering too many ifs in this file, what about concentrating all the instructions for cors here?

Eg:

- place de imports here
- `INSTALLED_APPS.append('corsheaders')`
- `MIDDLEWARES.append(...)`

templates/openwisp2/settings.py
```diff
@@ -494,3 +494,37 @@
...
+from corsheaders.defaults import default_methods, default_headers
+
+INSTALLED_APPS.append('corsheaders')
+MIDDLEWARE.insert(2, 'corsheaders.middleware.CorsMiddleware')
+{% if django_cors.replace_https_referer %}
+MIDDLEWARE.insert(6, 'corsheaders.middleware.CorsPostCsrfMiddleware')
```
Aryamanz29, line 505: Acc to `django-cors-headers` : `CorsMiddleware` should be placed as high as possible, especially before any middleware that can generate responses, such as `Django’s CommonMiddleware.`

templates/openwisp2/settings.py
Aryamanz29: Yes, It would be better to place all settings related to `cors` at one place 👍

nemesisdesign (changes requested):

templates/openwisp2/settings.py
```diff
@@ -494,3 +494,37 @@
...
 {% endfor %}
+
+{% if django_cors.enabled %}
+
+from corsheaders.defaults import default_methods, default_headers
+
+INSTALLED_APPS.append('corsheaders')
+MIDDLEWARE.insert(MIDDLEWARE.index('django.middleware.common.CommonMiddleware'), 'corsheaders.middleware.CorsMiddleware')
+{% if django_cors.replace_https_referer %}
+MIDDLEWARE.insert(MIDDLEWARE.index('django.middleware.csrf.CsrfViewMiddleware') + 1, 'corsheaders.middleware.CorsPostCsrfMiddleware')
```
Aryamanz29, line 505: **+1** because `'corsheaders.middleware.CorsPostCsrfMiddleware'` must be placed after `'django.middleware.csrf.CsrfViewMiddleware'`

defaults/main.yml
```diff
@@ -177,3 +177,22 @@ cron_delete_old_users: "'hour': 0, 'minute': 10"
 cron_cleanup_stale_radacct: "'hour': 0, 'minute': 20"
 cron_delete_old_postauth: "'hour': 0, 'minute': 30"
 cron_delete_old_radacct: "'hour': 1, 'minute': 30"
+
+# cross-origin resource sharing (CORS) settings
+django_cors:
```
pandafy, line 182: I don't like that we are creating an abstractions for the settings are provided by django-cors-headers. I believe this is bloating the ansible role and making it complex for very small benefits. @nemesisdesign are you sure we want to proceed this way? 

If we really want to add variables to control these settings, we should add them only for the required settings (settings that do not have any default). For rest, we should ask users to add the django setting using the `openwisp2_extra_django_settings_instructions` variable.

nemesisdesign: That could work if we explain it properly in the README.

pandafy:
Hey @Aryamanz29! Thank you for getting this work started. I will take it on from here. 😄

nemesisdesign (approved):

PR #392, opened 2022-08-18
[fix] Implemeted efficient reloading of supervisor services #365

### LINKED ISSUE #365: [bug] Supervisor reload takes too long
Reloading supervisor takes an excessive long time, I wonder if we can implement a more graceful reloading, [uwsgi supports reloading with zero downtime by using the SIGHUP signal](https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html), we should investigate whether SIGHUP would reload also the other processes or not, if yes we can use just SIGHUP for everything, otherwise we can use SIGHUP only for uwsgi but other services can restart normally.

Right now we are restarting the whole supervisor service which can take up to a few minutes to complete, too slow and not suited for production.

I think we could switch from using "reload" to "update all" as suggested by @pandafy (please double check that) for any change to any of the service definitions handled by supervisor.

Then for any application related change (python dependency upgrades, django settings changes, etc.) we could add a new handler which sends the SIGHUP signal.

pandafy:
Closes #365

### DIFF
--- a/README.md
+++ b/README.md
@@ -945,6 +945,11 @@ Below are listed all the variables you can customize (you may also want to take 
     openwisp2_admin_allowed_network: null
     # install ntp client (enabled by default)
     openwisp2_install_ntp: true
+    # if you have any custom supervisor service, you can
+    # configure it to restart along with other supervisor services
+    openwisp2_extra_supervisor_restart:
+        - name: my_custom_service
+          when: my_custom_service_enabled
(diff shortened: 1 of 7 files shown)


nemesisdesign (dismissed):
Looks good, let's keep testing before merging.

nemesisdesign (approved):

PR #395, opened 2022-09-02
[chore] Task name, variable and test changes for linting

broadstack-au:
In the interest of reducing the amount of noise generated during role tests, all bar one item that triggered linting errors have been silenced.

The remaining warning is:
```
ignore-errors: Use failed_when and specify error conditions instead of using ignore_errors.
tasks/apt.yml:51 Task/Handler: set redis to listen on ipv4
```
which I left alone because I have no context around how to trigger it / when it would error / why this ignore was introduced.

If you'd prefer this change to happen differently, please let me know.


nemesisdesign (approved):

nemesisdesign:
Thanks @broadstack-au :+1:

PR #402, opened 2022-09-22
[change] Use freeradius 3.2 on Ubuntu 22.04 #401

### LINKED ISSUE #401: [enhancement] Use FreeRADIUS 3.2 for installation
The role installs FreeRADIUS 3.0.x which fails on Ubuntu 22.04.1 because `libjson-c4` package is not available. 

Moreover, the [FreeRADIUS documentation](https://networkradius.com/packages/#fr30-ubuntu) does not mention support for Ubuntu 22.04 for FreeRADIUS 3.0. 

We should update the role to install FreeRADIUS 3.2

Aryamanz29:
Closes #401

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -149,7 +149,7 @@ openwisp2_radius_delete_old_postauth: 365
 openwisp2_radius_delete_old_radacct: 365
 openwisp2_radius_allowed_hosts: ["127.0.0.1"]
 openwisp2_freeradius_install: true
-freeradius_dir: "{{ '/etc/freeradius/3.0' if ansible_distribution_release | string == 'jammy' else '/etc/freeradius' }}"
+freeradius_dir: /etc/freeradius
 freeradius_mods_available_dir: "{{ freeradius_dir }}/mods-available"
 freeradius_mods_enabled_dir: "{{ freeradius_dir }}/mods-enabled"
 freeradius_mods_config_dir: "{{ freeradius_dir }}/mods-config"
--- a/tasks/freeradius.yml
+++ b/tasks/freeradius.yml
@@ -1,4 +1,15 @@
 ---
+- name: Add freeradius repository into Ubuntu jammy sources list
+  shell: |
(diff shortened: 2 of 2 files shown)


pandafy (changes requested):
@Aryamanz29 I have concerns on how these changes would affect an existing installation of FreeRADIUS on Ubuntu 22.04. I suspect that running these changes against an existing installation would result in installing FreeRADIUS 3.2 alongside FreeRADIUS 3.0. 

Also, changing `freeradius_dir` variable like this would be a breaking change to configuration of existing installations.

We have to be more careful with change. I have these suggestion for the flow

- If the OS is Ubuntu 22.04.1 (jammy), check if freeradius is already installed. We can check this with  `freeradius -v | sed -n 2p` which also gives us the version 
    - If the version of 3.0.x, do not change `freeradius_dir` variable, else update the variable to `/etc/freeradius` 
- If the OS is Ubuntu 22,04.1 (jammy) and FreeRADIUS is not installed (new installation), then update the `freeradius_dir` variable.

pandafy:
Hi @Aryamanz29! I will be taking over this PR. Can you please push any changes you have locally?

pandafy:
@nemesisdesign the builds are failing because `importlib-metadata` is incompatible with Python 3.7.  I have already pinned the version in [openwissp-utils](https://github.com/openwisp/openwisp-utils/blob/163b4d4aae3c1a0aa5b58bf612fe081dba812922/setup.py#L49-L51). Should I update the role to use the development version of openwisp-utils?

nemesisdesign (commented):
> @nemesisdesign the builds are failing because `importlib-metadata` is incompatible with Python 3.7. I have already pinned the version in [openwissp-utils](https://github.com/openwisp/openwisp-utils/blob/163b4d4aae3c1a0aa5b58bf612fe081dba812922/setup.py#L49-L51). Should I update the role to use the development version of openwisp-utils?

Let's release a new bugfix version of openwisp-utils with that fix.

nemesisdesign (approved):

PR #404, opened 2022-09-28
[feature] Added task add inventory_hostname to /etc/hosts

pandafy:
The /etc/hosts files is updated to resolve inventory_hostname to localhost (127.0.0.1).

**Blockers**
- [x] https://github.com/openwisp/ansible-openwisp2/pull/402

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -12,6 +12,10 @@
 - import_tasks: ssh.yml
   tags: [openwisp2, ssh]
 
+- name: Import system tasks
+  import_tasks: system.yml
+  tags: [openwisp2, system]
+
 - import_tasks: pip.yml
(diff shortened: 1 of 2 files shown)


tasks/main.yml
```diff
@@ -12,6 +12,9 @@
 - import_tasks: ssh.yml
   tags: [openwisp2, ssh]
 
+- import_tasks: system.yml
```
nemesisdesign, line 15: Can you fix the linter warnings that appear after adding this?
Later on we can fix all the other remaining issues too (I will open an issue).

nemesisdesign (changes requested):

nemesisdesign (approved):

Issue #400, opened 2022-09-13
error install source deadsnakes on bullseye


mikysal78:
https://github.com/openwisp/ansible-openwisp2/blob/4f0a14825ccc7abfd9727391a0b2b9027a711d26/tasks/apt.yml#L143

i have this error:
```
ASK [openwisp.openwisp2: Add deadsnakes PPA] **************************************************************************************************************************************************************************************
Tuesday 13 September 2022  23:07:37 +0200 (0:00:03.758)       0:02:51.539 ***** 
fatal: [controller.nnxx.ninux.org]: FAILED! => changed=false 
  msg: 'Failed to update apt cache: E:The repository ''http://ppa.launchpad.net/deadsnakes/ppa/ubuntu bullseye Release'' does not have a Release file.'

PLAY RECAP ******************************************************************************************************************************************************************************************************************************
controller.nnxx.ninux.org  : ok=46   changed=0    unreachable=0    failed=1    skipped=51   rescued=0    ignored=0   
```

emsu-io-suefke:
Pull request to fix this issue: https://github.com/openwisp/ansible-openwisp2/pull/407

mikysal78:
> Pull request to fix this issue: #407

When will the merge be done?
As of today, if you don't have ubuntu, you can't upgrade or install!

mikysal78:
I tried after the merge but I get the same error.
On my notebook i have os debian11.5 and i have this version of ansible:
```
michele@mikytux:/git/new-server$ ansible --version
ansible [core 2.12.9]
  config file = /git/new-server/ansible.cfg
  configured module search path = ['/home/michele/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /git/new-server/collections
  executable location = /usr/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 3.0.3
  libyaml = True
  ```
  my ansible.cfg is:
  ```
  michele@mikytux:/git/new-server$ more ansible.cfg 
[defaults]
module_set_locale=True
module_lang=it_IT.UTF-8
remote_port    = 2400
retry_files_enabled = False
nocows = 1

roles_path = roles
collections_paths = collections


ansible_managed = Ansible managed: modified on %d-%m-%Y

# Use the YAML callback plugin.
[...]

PR #407, opened 2022-10-06
Fix checking for Python version <3.7

emsu-io-suefke:
Running
$ ansible-playbook -i hosts playbook.yml -u user --check would always result in an error because the variable {{openwisp2_installed_python}} would never have been set when running with --check.

It is Ansible best practice to use Ansible native checks wherever possible and it is reasonable to assume that Ansible runs in the Python version the system has. For Python version checking, see https://serverfault.com/a/858684

Another change is to install >= Python 3.7 only if < Python 3.7 is found. The old expression looks as if it would install deadsnakes-python3.7 over system python3.7

Related to issue https://github.com/openwisp/ansible-openwisp2/issues/400

### DIFF
--- a/tasks/apt.yml
+++ b/tasks/apt.yml
@@ -138,7 +138,7 @@
     - name: Check openwisp2 should install python 3.7
       set_fact:
         openwisp2_should_install_python_37: |
-          {{ openwisp2_installed_python.stdout.split(' ')[1] is version_compare('3.7', 'le') }}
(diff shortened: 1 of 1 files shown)


mikysal78:
When will the merge be done?
As of today, if you don't have ubuntu, you can't upgrade or install!

nemesisdesign (approved):
Thanks! :+1:

PR #408, opened 2022-10-11
[feature] Added variable to configure FreeRADIUS listen address

pandafy:


templates/freeradius/openwisp_site.j2
```diff
@@ -1,7 +1,7 @@
 server default {
     listen {
         type = auth
-        ipaddr = *
+        ipaddr = {{ freeradius_openwisp_site_listen_ipaddr}}
```
nemesisdesign, line 4:
```suggestion
        ipaddr = {{ freeradius_openwisp_site_listen_ipaddr }}
```

Below too, what if we want to have both ipv4 and ipv6? Is that doable via this config var? Please double check.

nemesisdesign (changes requested):

templates/freeradius/openwisp_site.j2
pandafy: @nemesisdesign we can use only one ip address  per `listen` block. I tested locally with `::/128` and it worked as expected.

nemesisdesign (approved):

PR #412, opened 2022-11-21
[feature] Added ansible variable to allow configuring uwsgi parameters

pandafy:
Added "openwisp2_uwsgi_extra_conf" variable which allows appeding extra configuration for uwsgi in uwsgi.ini


nemesisdesign (approved):
Thanks! :+1:

PR #432, opened 2023-04-01
[bug] Updated README.md file

keshav-04:
username@IP_address for hosts updated in README.md file.
For the `ssh` command, we use `ssh username@IP_address` to access the remote desktop machine.
Also, in the hosts file given, we will use `[openwisp2] username@IP_address` to access the remote desktop machine via hosts file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,7 +390,7 @@ Create an ansible inventory file named `hosts` **in your working directory**
 [openwisp2]
 username@192.168.56.2
 ```
-  Replace username with the username of your VM.
+  Replace `username` with the username of your VM.
 
 **Step 5**: Create the ansible playbook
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -387,8 +388,9 @@ Create an ansible inventory file named `hosts` **in your working directory**
 
 ```
 [openwisp2]
-192.168.56.2
+username@192.168.56.2
```
nemesisdesign, line 391: this is not right

nemesisdesign (changes requested):

README.md
keshav-04: It is correct if we have only one user on the remote host, but if we have multiple hosts then to use the `192.168.56.2` syntax in our inventory file, we will need to specify the SSH user in our Ansible playbook.

PR #438, opened 2023-06-07
[docs] Add ansible.posix as dependency

minhng99:
Dependency of openwisp.influxdb/tasks/udp_mode.yml

```
ERROR! couldn't resolve module/action 'ansible.posix.sysctl'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in '/home/minhnq31/.ansible/roles/openwisp.influxdb/tasks/udp_mode.yml': line 1, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Set net.core.rmem_max
  ^ here
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,10 @@ collection installed
 
     ansible-galaxy collection install "community.general:>=3.6.0"
 
+Ensure that you have `ansible.posix` installed
+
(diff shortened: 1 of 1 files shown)


nemesisdesign (approved):
Thanks

Issue #398, opened 2022-09-12
error during install freeradius-role


chaosenze:
Hi,

it looks like it's the wrong path to freeradius config dir.

```
TASK [openwisp.openwisp2 : Enable SQL module] ************************************************************************************************************************************************************************************************
fatal: [openwisp2.dungeon.myhome]: FAILED! => {"changed": false, "msg": "Error while linking: [Errno 2] No such file or directory: b'/etc/freeradius/mods-available/sql' -> b'/etc/freeradius/mods-enabled/sql'", "path": "/etc/freeradius/mods-enabled/sql"}
```

```
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

```
$ ls -l /etc/freeradius/
total 0
drwxr-xr-x 1 freerad freerad 408 Sep 13 01:18 3.0
```

PR #461, opened 2024-01-17
[fix] Corrected FreeRADIUS directory

dim-0:
Modified freeradius_dir to include the `3.0/` directory


lukas-mika-maersk:
Fixes #398

pandafy:
I think this is already handled here 

https://github.com/openwisp/ansible-openwisp2/blob/ce426d33759db9e3d9ea5048521c226676473d88/tasks/freeradius.yml#L2-L32

nemesifier:
This change as is can't be accepted because it causes issues to systems which are using newer versions of freeradius.
I am closing for now. Feel free to reopen supplying a patch that can work in all the supported cases.
Thank you for contributing.

PR #471, opened 2024-05-08
[feature] Allowed deploying WPA Enterprise 2 TTLS-PAP #462

### LINKED ISSUE #462: [feature] Allow deploying WPA Enterprise 2 TTLS-PAP
Let's find a way to allow users to deploy [WPA Enterprise 2 TTLS-PAP via OpenWISP RADIUS](https://openwisp-radius.readthedocs.io/en/stable/developer/freeradius_wpa_enterprise.html) more easily via this role.

I think we should generate the sites automatically given a list of organizations, each org supplying UUID, secret and a name (used just for internal purposes). However, I would leave the option to override the generation of a site if the user supplies a path to a template, so they can completely customize the site if needed.

Eg:

```
freeradius_eap_orgs:
  - name: org_a
    uuid: ...
    secret: ...
  - name: org_b
    uuid: ...
    secret: ...
    template: <path>
```

pandafy:
Closes #462

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,6 +607,11 @@ installs FreeRADIUS, and configures it for WPA Enterprise (EAP-TTLS-PAP):
           acct_port: 1823
           # Port used by the authentication service of inner tunnel for this FreeRADIUS site
           inner_tunnel_auth_port: 18230
+          # If you want to use a custom certificate for FreeRADIUS EAP module,
+          # you can specify the path to the certificate and private key as follows.
+          # Ensure that the certificate and private key can be read by the "freerad" user.
+          cert: /etc/freeradius/certs/cert.pem
+          private_key: /etc/freeradius/certs/key.pem
         # You can add as many organizations as you want
         - name: demo
           uuid: 00000000-0000-0000-0000-000000000001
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -561,6 +561,153 @@ When the playbook is done running, if you got no errors you can login at:
...
+authentication, respectively. These custom ports are specified because the
+Ansible role creates a common FreeRADIUS site for all organizations, which
+also supports captive portal functionality. This common site is configured
+to listen on the default FreeRADIUS ports 1812, 1813, and 18120. Therefore, when
+configuring WPA Enterprise authentication for each organization, unique
+ports must be provided to ensure proper isolation and functionality.
+
+#### Using Let's Encrypt Certificate for WPA Enterprise (EAP-TTLS-PAP)
+
+In this section, we demonstrate how to utilize Let's Encrypt certificates for WPA Enterprise (EAP-TTLS-PAP) authentication. Similar to the [Automatic SSL certificate](#automatic-ssl-certificate), we use [geerlingguy.certbot](https://galaxy.ansible.com/geerlingguy/certbot/) role to automatically install and renew a valid SSL certificate.
```
nemesifier, line 637: Please wrap long lines.

tasks/freeradius.yml
```diff
@@ -200,3 +200,14 @@
     owner: freerad
     group: freerad
   notify: Restart freeradius
+
+- debug:
+    msg: "{{ freeradius_eap_orgs }}"
```
nemesifier, line 205: Is this intentionally left here?

nemesifier (changes requested):
Good progress @pandafy, can you add this functionality to the test playbook so we can have some basic test coverage for it?

Please also see my comments below.

nemesifier (changes requested):
One problem found during testing: the calling station ID will be the wifi interface, which is not the same of the mac address of the device. This will make it hard to connect the device to the radius accounting session.

Can you try changing the mac address of the wifi interface by using the `{{mac_address}}` variable?

I just tried this and it worked (I had to bring down and up the interface manually, openwrt did not update the mac address on its own).

Screenshot from 2024-05-29 19-51-39

Screenshot from 2024-05-29 19-51-31

It would be great to recommend this in the tutorial.

nemesifier (approved):

Issue #201, opened 2020-08-19
[feature] Automatically create management VPN


nemesisdesign:
This feature can be enabled by default, users should be able to disable it with a configuration variable, we should guide users to do this choice in the beginning of the tutorial.

If enabled, we should install openvpn, we should create a VPN server object (if no VPN server object exist yet), we should place a default working configuration for a layer 3 VPN (routed), then create a VPN-client template based on it and flag this template as default.
We should then take the openvpn server configuration just saved into OpenWISP and deploy it in the linux filesystem, enable the openvpn service and start it.

This way we'll get a basic management VPN working out of the box for new installations, which coupled with other features (#195 #199 #200) will make this role a lot more powerful and helpful for the community.

stktyagi:
Hi,

Should the OpenVPN service be restarted on every playbook run, or only when config changes?

Riya61018:
I will Start working on this issue .

nemesifier:
@Riya61018 I advise to not do that.

We need a comprehensive solution which can handle configuration updates, certificate revocation lists, see [VPN Deployer](https://openwisp.io/docs/dev/developer/gsoc-ideas-2026.html#openwisp-vpn-deployer-linux-package), I believe @asmodehn is aware of this, we'll surely need to discuss more.

But adding the simple creation of an OpenVPN server in this role is not going to be a definitive solution and is also going to add more complexity to this role. It sounds like a good idea but as the amount of VPNs we support grows, it becomes more and more clear that we need a tool like the [VPN Deployer](https://openwisp.io/docs/dev/developer/gsoc-ideas-2026.html#openwisp-vpn-deployer-linux-package).

Riya61018:
ok
i will find something else to solve

PR #562, opened 2025-11-14
Vpn management creation

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.
- [ ] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [ ] I have updated the documentation.

First Draft for implementing the server part of #201.
Only tested in molecule for now.

## Description of Changes

- If `openwisp2_openvpn_setup` is set to true
- If there is no VPN objects in openwisp
A VPN server object is created by the load_initial_data.py script.
Its name and id are known by ansible only for this run, which makes it possible to download the configuration and start the openvpn server.

That is if there is no error...

If an error occurs somewhere after adding the VPN object in the DB, the name and id of the VPN will not be available on the following runs, which means the setup must be completed manually...
Not sure if this is acceptable and/or if there is a better way to address that problem (known unique vpn_server_name that we can look for on every run instead ?)

In any case, let me know what you think about this...

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -234,3 +234,5 @@ openwisp2_radius_periodic_tasks: true
 openwisp2_usage_metric_collection_periodic_tasks: true
 # point {{ inventory_name }} to localhost in /etc/hosts
 openwisp2_inventory_hostname_localhost: true
+# installs and configure openvpn to provide a Management IP to all devices
+openwisp2_openvpn_setup: true
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,0 +1,4 @@
+---
+requires_ansible: ">=2.13"
+collections:
+  - community.general
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -28,7 +28,7 @@ provisioner:
     verify: ../resources/verify.yml
   config_options:
     defaults:
-      stdout_callback: yaml
+      result_format: yaml
       bin_ansible_callbacks: true
   inventory:
     host_vars:
--- a/molecule/local/molecule.yml
+++ b/molecule/local/molecule.yml
@@ -4,6 +4,7 @@ dependency:
   name: galaxy
   options:
     role-file: molecule/resources/requirements.yml
+    requirements-file: molecule/resources/collections.yml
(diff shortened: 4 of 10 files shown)


DominatrixiaALLATU (approved):

DominatrixiaALLATU (changes requested):
Maybe

nemesifier (commented):
Thanks for tackling #201, this is a genuinely useful direction and I appreciate that you flagged the fragility yourself in the description. Let me answer the design question directly because it determines the whole approach.

**The core design problem you raised is the blocker**

Capturing `openwisp2_vpn_id` / `openwisp2_vpn_name` by regex-parsing the `load_initial_data.py` stdout and only on the run that prints `created` is not robust:

- If anything after the DB insert fails, the id/name are lost and every subsequent run skips the block, so the server is left half-configured with no automatic recovery.
- It is also not idempotent: re-running the role on a healthy install does nothing useful because the facts are only set on the creating run.

Your own suggestion is the right one: use a well-known, fixed VPN name (for example derived from a role variable), and on every run look the VPN up by that name via the management script or API instead of parsing creation output. That makes the play idempotent and self-healing: "ensure a VPN named X exists, then fetch its id, then configure openvpn". Please pursue that path; the stdout-regex approach should be dropped.

**Other issues to fix once the approach is settled**

1. `openwisp2_openvpn_setup` defaults to `true` in `defaults/main.yml`. A new feature that installs and starts OpenVPN must default to `false`, otherwise every existing deployment that re-runs the role suddenly gets an OpenVPN server. Opt-in only.
2. Missing trailing newline in `defaults/main.yml` and `molecule/resources/collections.yml` (yamllint will complain).
3. The regex for `openwisp2_vpn_name` is greedy to end of line and will capture trailing whitespace / CR; even in a corrected design, parsing names this way is brittle.
4. Needs real tests and docs before merge (both unchecked, which is honest, just noting it for the final pass).

Good initiative. Once you switch to "lookup by fixed name, idempotent" and make it opt-in, I am happy to review the next iteration in detail.

PR #575, opened 2025-12-04
[change] Added openwisp2_pip_extra_args: "--upgrade" to force pip to upgrade deps

### LINKED ISSUE #572: [feature] Ensure all Python dependencies update on existing installs when playbook runs
**Is your feature request related to a problem? Please describe.**
Currently, running the playbook on an existing installation may not update all Python dependencies. This can lead to issues if packages are outdated.

**Describe the solution you'd like**
Update the tasks so that Python dependencies installed with pip) are always updated (force reinstalled) when the playbook is rerun on an existing setup. This ensures environments are consistent and all dependencies are brought up-to-date automatically.

Piyushrathoree:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [x] I have manually tested the changes proposed in this pull request.
- [ ] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [ ] I have updated the documentation.

Closes #572 

## Description of Changes

update the pip.yml file with extra-args:"--upgrade" which make sure that every dependency is updated.


Piyushrathoree:
@pandafy PTAL

pandafy (changes requested):
Good start @Piyushrathoree! 

I am afraid that `--upgrade` might not always work when we use tarball URL as source. E.g. when we set openwisp-controller version in the ansible playbook 

```yaml
openwisp2_controller_version: openwisp-controller @ https://github.com/openwisp/openwisp-controller/tarball/master
``` 

In such scenarios, we would need to use [`state: forcereinstall`](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/pip_module.html#parameter-state). 

Using `forcereinstall` by default will slow down the playbook, since it will make pip reinstall packages even when the latest package have been installed. 

I can think of exposing `pip.extra_args` argument through ansible variable `openwisp2_pip_extra_args`. The default value for this argument will be `openwisp2_pip_extra_args: "--upgrade"`. When needed, the user can override the variable in the playbook to use `openwisp2_pip_extra_args: "--upgrade --force-reinstall" which will force-reinstall the packages. 

We will need to add this new variable in https://github.com/openwisp/ansible-openwisp2/blob/master/docs/user/role-variables.rst, and explain the usage with a comment. You can take reference from other settings mentioned in the file.

pandafy (changes requested):

nemesifier (approved):

PR #601, opened 2026-02-14
[feature] Configured nginx to serve precompressed files (gzip/brotli) #599

### LINKED ISSUE #599: [feature] Configure nginx to serve precompressed files (gzip/brotli)
Depends on https://github.com/openwisp/openwisp-utils/issues/565 being resolved.

Let's update ansible-openwisp2 to remove the settings which turned off GZIP and BROTLI compression, also because they won't be valid anymore with the new package we're migrating to.

Let's then configure nginx to serve precompressed files (both  gzip/brotli).

SAMurai-16:
Remove GZIP_STATIC_COMPRESSION and BROTLI_STATIC_COMPRESSION settings from Django configuration as they are no longer valid with the new openwisp-utils package. Configure nginx to serve precompressed files by adding gzip_static and brotli_static directives to the SSL configuration. This improves performance by serving pre-compressed .gz and .br files directly when available, rather than compressing on-the-fly.

Updated templates/openwisp2/settings.py to remove obsolete compression settings, modified defaults/main.yml to add precompressed file serving configuration, and updated documentation in docs/user/role-variables.rst.

Fixes #599

## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [x] I have manually tested the changes proposed in this pull request.
- [x] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [x] I have updated the documentation.

Closes #599 .


nemesifier (commented):
@SAMurai-16 when the CI build does not fail due to your changes, the best thing is to:

1. don't fight it in this PR, do your changes as if the CI Build is not failing
2. open another PR which aims to A) Demonstrate the build fails on master, B) Help to debug it and fix it is always welcome.

nemesifier (commented):
It looks pretty good! I'll do some testing as soon as I can to confirm all works as expected and then merge. Thanks!

defaults/main.yml
```diff
@@ -161,6 +161,9 @@ openwisp2_nginx_ssl_config:
     - "application/xml"
     - "application/x-font-ttf"
     - "font/opentype"
+# Precompressed files serving: 'auto' (default), true, or false
+openwisp2_nginx_gzip_static: auto
+openwisp2_nginx_brotli_static: auto
```
nemesifier: These keys correctly sit outside `openwisp2_nginx_ssl_config`, but wedged between `gzip_types` items and `openwisp2_nginx_access_log` they look like they belong to the dict. Move them below `openwisp2_nginx_error_log` with a blank line above.

tasks/apt.yml
```diff
@@ -39,6 +39,34 @@
...
 
+- name: Check if nginx gzip_static module is available
+  shell: "nginx -V 2>&1 | grep -q 'http_gzip_static_module' && echo 'available' || echo 'not_available'"
+  register: gzip_static_check
+  when: openwisp2_nginx_install
+  changed_when: false
+
+- name: Check if nginx brotli module is available
+  shell: "nginx -V 2>&1 | grep -q 'brotli' && echo 'available' || echo 'not_available'"
+  register: brotli_check
```
nemesifier: Use `ansible.builtin.command: nginx -V` once (the rest of the file uses the FQCN), register the result, and derive both facts in Jinja from `stderr`, `nginx -V` prints to stderr. The current `grep -q … && echo … || echo …` always exits 0, so any genuine failure is silently masked.

tasks/apt.yml
```diff
@@ -39,6 +39,34 @@
...
   when: openwisp2_nginx_install
 
+- name: Check if nginx gzip_static module is available
+  shell: "nginx -V 2>&1 | grep -q 'http_gzip_static_module' && echo 'available' || echo 'not_available'"
+  register: gzip_static_check
+  when: openwisp2_nginx_install
+  changed_when: false
+
+- name: Check if nginx brotli module is available
+  shell: "nginx -V 2>&1 | grep -q 'brotli' && echo 'available' || echo 'not_available'"
```
nemesifier: This won't detect brotli on Debian/Ubuntu, where it ships as a dynamic module (`libnginx-mod-http-brotli-filter`) loaded via `load_module` at runtime, `nginx -V` only lists compile-time modules. Use `nginx -T 2>&1 | grep -q brotli_filter_module` (loaded config) or probe the `.so`.

tasks/apt.yml
```diff
@@ -39,6 +39,34 @@
...
+  changed_when: false
+
+- name: Set nginx module availability facts
+  set_fact:
+    nginx_gzip_static_available: >-
+      {{ (gzip_static_check.stdout | default('not_available')) == 'available' and
+         (openwisp2_nginx_gzip_static == true or openwisp2_nginx_gzip_static == 'auto') }}
+    nginx_brotli_static_available: >-
+      {{ (brotli_check.stdout | default('not_available')) == 'available' and
+         (openwisp2_nginx_brotli_static == true or openwisp2_nginx_brotli_static == 'auto') }}
```
nemesifier: Two issues here:

1. `auto` and `true` behave identically, both gate on module availability. Either drop `auto` and use a plain boolean, or make `true` mean "required, fail if the module is missing". As is, `true` adds nothing over `auto`.
2. `== true` is fragile: a value coming in as the string `"true"` (e.g. `-e openwisp2_nginx_gzip_static=true`) won't match. Use `in [true, 'auto']` or cast with `| bool`.

tasks/apt.yml
```diff
@@ -39,6 +39,34 @@
...
+  when: openwisp2_nginx_install
+
+- name: Set nginx module availability facts (nginx not installed)
+  set_fact:
+    nginx_gzip_static_available: false
+    nginx_brotli_static_available: false
+  when: not openwisp2_nginx_install
```
nemesifier: These two `set_fact` blocks collapse into one task with a conditional expression.

docs/user/role-variables.rst
```diff
@@ -162,6 +162,16 @@ take a look at `the default values of these variables
...
+        # Precompressed gzip files serving control
+        # - 'auto' (default): automatically enabled when nginx gzip_static module is detected
+        # - true: enable only when nginx gzip_static module is available
+        # - false: explicitly disable even when module is available
+        openwisp2_nginx_gzip_static: auto
+        # Precompressed brotli files serving control
+        # - 'auto' (default): automatically enabled when nginx ngx_brotli module is detected
+        # - true: enable only when nginx ngx_brotli module is available
+        # - false: explicitly disable even when module is available
+        openwisp2_nginx_brotli_static: auto
```
nemesifier, line 182: Add a note that brotli is not shipped with stock nginx, on Debian/Ubuntu users need `libnginx-mod-http-brotli-filter` and `libnginx-mod-http-brotli-static`, otherwise `auto` does nothing.

molecule/resources/verify.yml
```diff
@@ -41,7 +41,7 @@
       command: systemctl status redis-server
       changed_when: false
 
-    - name: Chcke if redis is running
+    - name: Check if redis is running
```
nemesifier: Unrelated to this PR, split into its own commit.

molecule/resources/verify.yml
```diff
@@ -65,3 +65,43 @@
...
+        # Assert gzip_static directive is present when module is available
+        - name: Assert gzip_static directive is present (when module available)
+          shell: "grep 'gzip_static on;' /etc/nginx/sites-enabled/{{ inventory_hostname }}"
+          register: gzip_static_check
+          changed_when: false
+          when: nginx_gzip_module_check.stdout == 'available'
+
+        # Assert brotli_static directive is present when module is available
+        - name: Assert brotli_static directive is present (when module available)
+          shell: "grep 'brotli_static on;' /etc/nginx/sites-enabled/{{ inventory_hostname }}"
```
nemesifier: Trailing whitespace.

molecule/resources/verify.yml
```diff
@@ -65,3 +65,43 @@
...
+          changed_when: false
+          when: nginx_gzip_module_check.stdout == 'available'
+
+        # Assert brotli_static directive is present when module is available
+        - name: Assert brotli_static directive is present (when module available)
+          shell: "grep 'brotli_static on;' /etc/nginx/sites-enabled/{{ inventory_hostname }}" 
+          register: brotli_static_check
+          changed_when: false
+          when: nginx_brotli_module_check.stdout == 'available'
+
```
nemesifier: Only the positive case is asserted. Add the negative: directive absent when the module is missing, and when the variable is set to `false`.

nemesifier (changes requested):
Brotli detection is the critical one: on Debian/Ubuntu brotli ships as a dynamic module loaded via `load_module` at runtime, so `nginx -V` will not list it and `auto` ends up silently useless for most users. `auto` vs `true` semantics also need clarifying (right now they behave the same). Other comments inline.

SAMurai-16:
I'll work on the changes and will raise a PR soon

nemesifier (commented):
This should be ready, will give it a try asap.

nemesifier (dismissed):

nemesifier (approved):
@SAMurai-16 I made a number of improvements, feel free to see my changes. There were a number of problems:

1. nginx logic was placed in the wrong file, it belongs to nginx.yml
2. apt should be only concerned with installing system packages
3. the brotli file was not being installed, the probability that some user would install it manually is low, so the brotli feature would be practically unused
4. after testing on a VM with `strace` I verified that nginx wasn't serving the precompressed gzip files, it was instead serving the on-the-fly gzipped, I had to change the config slightly to make it prefer the pre-compressed gzip files
5. lots of logic, especially the molecule tests, was too verbose

I appreciate the effort, but there's no point in submitting PRs if you don't verify on your end that the proposed changes really work so I ask you to refrain from doing this.

Please read our [Anti AI Spam policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-spam-policy), with particular attention to the points which mention shifting the burder of testing on maintainers.

Thanks.

